PR #111, opened 2018-12-25
Support CSV

kennytm:

Implement CSV parsing.

1. A custom new CSV parser is introduced. It will parse every line of CSV and convert into the corresponding SQL row syntax. (The parser is implemented using ragel for simplicity. Probably a hand-written parser is faster but I haven't checked.)
2. The methods of the SQL parser used in `restore.go` is extracted into an `interface`.
3. If a data file's extension is `*.csv`, a CSV parser is created. Otherwise the normal SQL parser is created.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

 - Increased code complexity

Related changes


GregoryIan:
@WangXiangUSTC @lonng PTAL

lightning/mydump/csv_parser.go
```diff
@@ -0,0 +1,151 @@
...
+	if parser.pos == 0 && parser.cfg.Header {
+	outside:
+		for {
+			tok, content, err := parser.lex()
+			if err != nil {
+				return errors.Trace(err)
+			}
+			switch tok {
+			case csvTokSep:
+			case csvTokField:
```
lonng: Is it necessary to check the validity of the header? e.g: `col1,col2,,,col5`.

kennytm: We could; though it is probably unnecessary in this PR. With `col1,col2,,,col5`, this will produce either one of

```sql
INSERT INTO `t` (`col1`, `col2`, NULL, NULL, `col5`) VALUES ...;
INSERT INTO `t` (`col1`, `col2`, ``, ``, `col5`) VALUES ...;
```

and both will yield errors (syntax error or "unknown column" error)

Now, these *will* be a problem in #136 where we no longer assemble the SQL statement, but these functions will need to be rewritten anyway.

lightning/mydump/csv_parser.go
```diff
@@ -0,0 +1,151 @@
...
+
+func (parser *CSVParser) appendField(content []byte, shouldUnquote bool, quote byte) {
+	if !parser.cfg.BackslashEscape {
+		content = bytes.Replace(content, []byte(`\`), []byte(`\\`), -1)
+	}
+	if shouldUnquote {
+		content = bytes.Replace(content[1:len(content)-1], parser.delim2, parser.delim1, -1)
+	}
+	if !parser.cfg.NotNull && parser.cfg.Null == string(content) {
+		parser.rowBuf.WriteString(",NULL")
```
lonng:
```
# whether the CSV contains any NULL value
not-null = false
# if non-null = true, fields containing this value will be treated as NULL
null = '\N'
```
It seems that the logic and the configuration comment are not match.

lightning/config/config.go
```diff
@@ -97,13 +97,24 @@ type PostRestore struct {
 	Analyze       bool  `toml:"analyze" json:"analyze"`
 }
 
+type CSVConfig struct {
+	Separator       string `toml:"separator" json:"separator"`
+	Delimiter       string `toml:"delimiter" json:"delimiter"`
+	Header          bool   `toml:"header" json:"header"`
+	TrimLastSep     bool   `toml:"trim-last-separator" json:"trim-last-separator"`
+	NotNull         bool   `toml:"not-null" json:"not-null"`
+	Null            string `toml:"null" json:"null"`
```
GregoryIan: any comment for it？

GregoryIan:
LGTM

tidb-lightning.toml
```diff
@@ -80,6 +80,23 @@ no-schema=false
...
+[mydumper.csv]
+# separator between fields, should be an ASCII character.
+separator = ','
+# string delimiter, can either be an ASCII character or empty string.
+delimiter = '"'
+# whether the CSV files contain a header. If true, the first line will be skipped
+header = true
+# whether the CSV contains any NULL value
+not-null = false
+# if non-null = true, fields containing this value will be treated as NULL
```
lonng:
```suggestion
# if non-null = false, fields containing this value will be treated as NULL
```

lightning/mydump/csv_parser.go
```diff
@@ -0,0 +1,151 @@
...
+		tok, content, err := parser.lex()
+		switch errors.Cause(err) {
+		case nil:
+			break
+		case io.EOF:
+			if hasField {
+				tok = csvTokNewLine
+				break
+			}
+			fallthrough
```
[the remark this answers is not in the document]
kennytm: Yes it will return `io.EOF`, which would be treated specially. 

Added a test case to verify the expected behavior with an empty file, with or without headers.

kennytm:
PTAL @lonng @WangXiangUSTC

lonng:
LGTM

WangXiangUSTC:
LGTM

PR #132, opened 2019-02-16
restore: write index kv pairs and data kv pairs to separate engine

lonng:

Write index kvs and data kvs to separate engine. See: #137 

Use different engine file.

### Check List 

Tests 

 - Unit test
 - Integration test

Side effects

Related changes


lonng:
@kennytm PTAL

GregoryIan:
can we creare an issue for Proposal#4 ？

lonng:
@GregoryIan PTAL

lightning/restore/restore.go
```diff
@@ -339,19 +348,42 @@ func (rc *RestoreController) listenCheckpointUpdates(wg *sync.WaitGroup) {
...
 		// 	panic("forcing failure due to FailIfStatusBecomes")
 		// }
-		// continue
+		// goto gofailLabel2
+	gofailLabel2:
+		if false {
+			goto gofailLabel2
+		}
+
+		// gofail: var FailIfIndexEngineImported int
```
GregoryIan: it seems failpoint `FailIfStatusBecomes` is able to take place of `FailIfIndexEngineImported `

lightning/restore/restore.go
```diff
@@ -529,38 +597,50 @@ func (t *TableRestore) restoreTable(
...
 		common.AppLogger.Infof("[%s] import whole table takes %v", t.tableName, time.Since(timer))
 		err := engineErr.Get()
-		rc.saveStatusCheckpoint(t.tableName, -1, err, CheckpointStatusImported)
 		if err != nil {
 			return errors.Trace(err)
 		}
+
+		// If index engine file has been closed but not imported only if context cancel occurred
+		// when `importKV()` execution, so `UnsafeCloseEngine` and continue import it.
+		if indexEngineCp.Status == CheckpointStatusClosed {
```
GregoryIan: what if lightning execute L793 (L794 doesn't take effect), and then lightning was killed directly?

lightning/config/config.go
```diff
@@ -229,11 +230,17 @@ func (cfg *Config) Load() error {
...
-		*cfg.PostRestore.Level1Compact = true
+		*cfg.PostRestore.Level1Compact = false
 	}
 
+	if cfg.App.TableConcurrency < 2 {
+		cfg.App.TableConcurrency = 2
+	}
+	if cfg.App.IndexConcurrency < 2 {
+		cfg.App.IndexConcurrency = 2
+	}
```
kennytm: I don't think you need to mandate the minimal concurrency at all

lightning/restore/file_checkpoints.proto
```diff
@@ -28,12 +28,14 @@ message TableCheckpointModel {
...
     int64 alloc_base = 4;
     repeated EngineCheckpointModel engines = 6;
+    repeated uint32 index_engine_statuses = 7;
 }
 
 message EngineCheckpointModel {
     uint32 status = 1;
     // key is "$path:$offset"
     map<string, ChunkCheckpointModel> chunks = 2;
+    int32 engine_id = 3;
```
kennytm: You shouldn't need engine_id at all! In RFC 4 the status should be written into the `index_engine_statuses` field!

lightning/restore/restore.go
```diff
@@ -339,19 +348,42 @@ func (rc *RestoreController) listenCheckpointUpdates(wg *sync.WaitGroup) {
...
 		// if _, ok := scp.merger.(*ChunkCheckpointMerger); ok {
 		// 	wg.Wait()
 		// 	panic("forcing failure due to FailIfImportedChunk")
 		// }
-		// continue
+		// goto gofailLabel1
+	gofailLabel1:
+		if false {
+			goto gofailLabel1
+		}
```
kennytm: You don't need these `if false` stuff, see the function `OpenEngine` in `kv/importer.go`.

lightning/config/config.go
lonng: adressed

lightning/restore/checkpoints.go
```diff
@@ -108,8 +109,9 @@ type ChunkCheckpoint struct {
...
 
 type EngineCheckpoint struct {
-	Status CheckpointStatus
-	Chunks []*ChunkCheckpoint // a sorted array
+	EngineID int
+	Status   CheckpointStatus
+	Chunks   []*ChunkCheckpoint // a sorted array
 }
 
 type TableCheckpoint struct {
```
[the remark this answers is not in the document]
lonng: Actually, we should treat it as a normal Engine.

lightning/restore/checkpoints.go
```diff
@@ -192,7 +194,7 @@ func (merger *StatusCheckpointMerger) SetInvalid() {
 }
 
 func (merger *StatusCheckpointMerger) MergeInto(cpd *TableCheckpointDiff) {
-	if merger.EngineID == -1 || merger.Status <= CheckpointStatusMaxInvalid {
+	if merger.EngineID == invalidEngineID || merger.Status <= CheckpointStatusMaxInvalid {
```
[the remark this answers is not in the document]
lonng: addressed

lightning/restore/restore.go
lonng: We need an additional failpoint to cover FailIfIndexEngineImported also. See `tests/checkpoint`

lightning/restore/restore.go
lonng: addressed

lightning/restore/restore.go
lonng: addressed

lightning/restore/restore.go
```diff
@@ -54,6 +56,11 @@ const (
 	defaultGCLifeTime = 100 * time.Hour
 )
 
+const (
+	indexEngineID   = -1
+	invalidEngineID = math.MinInt64
```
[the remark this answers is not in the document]
lonng: addressed

lightning/restore/restore.go
```diff
@@ -501,14 +533,45 @@ func (t *TableRestore) restoreTable(
...
+			defer rc.indexWorkers.Recycle(indexWorker)
+			var err error
+			indexEngine, err = rc.importer.OpenEngine(ctx, t.tableName, indexEngineID)
+			if err != nil {
+				return errors.Trace(err)
+			}
+		}
 
+		// The table checkpoint status less than `CheckpointStatusIndexImported` implies
+		// that index engine checkpoint status less than `CheckpointStatusImported`.
```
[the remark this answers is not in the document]
lonng: addressed

GregoryIan:
LGTM

lightning/config/config.go
```diff
@@ -229,11 +231,10 @@ func (cfg *Config) Load() error {
 		}
 	}
 
-	// If the level 1 compact configuration not found, default to true
+	// If the level 1 compact configuration not found, default to false
```
kennytm: If we're changing the default to `false` it doesn't need the pointer indirection anymore (but why changing it to `false` in this PR?)

lightning/restore/file_checkpoints.proto
kennytm: Let's discuss about this tomorrow morning

PR #136, opened 2019-03-03
[WIP] Skip the entire KV encoder

kennytm:
### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Side effects

 - Increased code complexity

Related changes


kennytm:
After this PR, the time to read + encode + deliver the entire 256 MiB file of the 6.5 TiB schema is reduced from ~3.75 minutes to ~2.5 minutes.

But now we need to clear the review of all other PRs first 🤷...

lonng:
We need to take account in that values contain an expressions, e.g. `insert into t1 (c1, c2, c3) values (100, default(c2), year('2018-02-03'))`.

I think we can reuse `pingcap/parser` to parse `content` to `ast.ExprNode` and other API(https://github.com/pingcap/tidb/blob/0b371ea4fa95a829aa83376d9882dae10bee15d2/expression/expression.go#L39) to get the `types.Datum`.

PR #141, opened 2019-03-14
*: parameter type fix

amyangfei:
fix error
```bash
cmd/tidb-lightning-ctl/main.go:257:40: cannot use engineID (type int) as type int32 in argument to importer.UnsafeCloseEngine
```

add type cast

### Check List 

Tests 

 - Unit test
 - Integration test


amyangfei:
PTAL @kennytm @lonng

PR #145, opened 2019-03-14
*: parse the data source directly into data and skip the KV encoder

kennytm:

In our testing with a 4.1 TB workload, we found that parsing SQL takes almost half of the time to encode a row. Since we have already used a parser to extract each row, parsing it again is wasting computing resource. Additionally, for CSV we need to perform the complex and unnecessary Parse CSV → Reconstruct SQL → Parse SQL.

We change the Lightning parsers to directly produce an array of `types.Datum` for both CSV and SQL. We also get rid of the abstraction layer `KvEncoder` (since it only accepts SQL statements), and directly use `(*table.Table).AddRecord` to convert the `[]types.Datum` into KV pairs. 

This slashes half of the encoding time according to experiment. 

### Check List 

Tests 

 - Unit test
 - Integration test

Side effects

Related changes


kennytm:
@lonng  Some metrics are temporarily removed, we need to see if we want to tweak the metrics or the process. The old process:

* Read a block (64 KB) of SQL values → Reconstruct INSERT statement → Execute INSERT statement to get KV pairs → Deliver block

New process:

* Read a *single row* and copy to datum array → Encode to get KV pairs of this row → Buffer until 64 KB → Deliver buffer

The following metrics were involved in this change and may need to be repurposed?

* ChunkParserReadRowSecondsHistogram 
* BlockReadSecondsHistogram (concept of "block" no longer applies)
* BlockReadBytesHistogram 
* BlockEncodeSecondsHistogram

lonng:
@kennytm I think we could add some new metrics about:

- Encode to get KV pairs duration → data part + index part
- Data KV size
- Index KV size

lightning/kv/session.go
```diff
@@ -0,0 +1,225 @@
...
+	"github.com/pingcap/tidb/util"
+	"github.com/pingcap/tidb/util/kvcache"
+	kvec "github.com/pingcap/tidb/util/kvencoder"
+	binlog "github.com/pingcap/tipb/go-binlog"
+)
+
+// transaction is a trimmed down Transaction type which only supports adding a
+// new KV pair.
+type transaction struct {
+	kvPairs []kvec.KvPair
```
lonng: It is better to compound `kv.Transaction` interface here to avoid `panic("unexpected ***() call")`.

kennytm:
/run-all-tests

> Post https://coveralls.io/api/v1/jobs: net/http: TLS handshake timeout
>
> — https://internal.pingcap.net/idc-jenkins/blue/organizations/jenkins/lightning_ghpr_test/detail/lightning_ghpr_test/86/pipeline

kennytm:
PTAL @lonng

lightning/common/util.go
```diff
@@ -181,9 +182,27 @@ func IsRetryableError(err error) bool {
...
-// cancellation. This function returns `false` (not a context-canceled error) if
-// `err == nil`.
+// cancellation. This function should only be used for inhabiting logs related
+// to canceling, where the log is usually just noise.
+//
+// This function returns `false` when the current logging level is set to
+// "debug".
+//
+// This function returns `false` (not a context-canceled error) if `err == nil`.
 func IsContextCanceledError(err error) bool {
```
lonng: I think we should rename this function base on its real meaning.

lightning/config/config.go
```diff
@@ -161,7 +164,8 @@ func NewConfig() *Config {
 			CheckRequirements: true,
 		},
 		TiDB: DBStore{
-			SQLMode:                    "STRICT_TRANS_TABLES,NO_ENGINE_SUBSTITUTION",
+			StrSQLMode:                 "STRICT_TRANS_TABLES,NO_ENGINE_SUBSTITUTION",
```
lonng: It is better to use `parser/mysql.DefaultSQLMode`.

lightning/config/config.go
```diff
@@ -226,6 +230,11 @@ func (cfg *Config) Load() error {
 		return errors.New("invalid config: `mydumper.csv.delimiter` must be one byte long or empty")
 	}
 
+	cfg.TiDB.SQLMode, err = mysql.GetSQLMode(cfg.TiDB.StrSQLMode)
```
lonng: `cfg.TiDB.SQLMode` will always assign here, so I think initialize `SQLMode` in L168 is redundancy.

lightning/config/config.go
```diff
@@ -226,6 +230,11 @@ func (cfg *Config) Load() error {
...
+	if err != nil {
+		return errors.New("invalid config: `mydumper.tidb.sql_mode` must be a valid SQL_MODE")
```
lonng:
```suggestion
		return errors.Annotate(err, "invalid config: `mydumper.tidb.sql_mode` must be a valid SQL_MODE")
```

lightning/metric/metric.go
```diff
@@ -169,18 +177,18 @@ var (
 
 func init() {
 	prometheus.MustRegister(IdleWorkersGauge)
-	prometheus.MustRegister(EngineCounter)
+	prometheus.MustRegister(ImporterEngineCounter)
```
lonng: `EngineCounter` was used in L213, but not registered here, is it expected?

lightning/kv/importer.go
```diff
@@ -203,6 +203,11 @@ func (importer *Importer) OpenEngine(
 	}, nil
 }
 
+// Tag gets an identification stirng of this engine for logging.
+func (engine *OpenedEngine) Tag() string {
```
lonng: Seems this method was never used, is it keep consistent with `ClosedEngine.Tag()`?

lightning/kv/session.go
```diff
@@ -0,0 +1,85 @@
...
+	vars.StmtCtx.TimeZone = vars.Location()
+	return &session{
+		txn:  transaction{},
+		vars: vars,
+	}
+}
+
+func (se *session) takeKvPairs() []kvec.KvPair {
+	pairs := se.txn.kvPairs
+	se.txn.kvPairs = make([]kvec.KvPair, 0, len(pairs))
```
lonng: Maybe we could use a memory pool to reduce memory allocation.

lightning/kv/sql2kv.go
```diff
@@ -15,79 +15,77 @@ package kv
...
+			value, err = types.NewIntDatum(rowID), nil
+		}
+		if err != nil {
+			return nil, errors.Trace(err)
+		}
+		record = append(record, value)
 	}
 
-	return kvPairs, rowsAffected, nil
+	_, err = kvcodec.tbl.AddRecord(kvcodec.se, record)
```
lonng: The `record` can be cached as a field of `TableKVEncoder` and reset it after call `kvcodec.tbl.AddRecord`.

lightning/restore/restore.go
```diff
@@ -1359,6 +1379,138 @@ func splitIntoDeliveryStreams(totalKVs []kvenc.KvPair, splitSize int) [][]kvenc.
...
+		return errors.Trace(err)
+	}
+
+	for _, kvs := range splitIntoDeliveryStreams(totalKVs, maxDeliverBytes) {
+		if err := stream.Put(kvs); err != nil {
+			return errors.Trace(err)
+		}
+	}
+
+	if err := stream.Close(); err != nil {
```
lonng: Stream will leak if error occurred in `stream.Put`

lightning/restore/restore.go
```diff
@@ -1367,154 +1519,31 @@ func (cr *chunkRestore) restore(
...
+		deliverCompleteCh <- deliverResult{dur, err}
 	}()
 
-	var buffer bytes.Buffer
+	defer close(kvsCh)
+
+	initializedColumns := false
+outside:
 	for {
 		select {
```
lonng: We should check `deliverCompleteCh` here, think about the below scenario:

1. `cr.deliverLoop` occurred an error in any branch L1453/L1461/L1463
2. will send a result `deliverResult{dur, err}` to `deliverCompleteCh`
3. nothing will be received from `kvsCh` because of `cr.deliverLoop` is returned
4. L1453 will be blocked if the channel is full.

Oops, dead lock.

lightning/restore/restore.go
```diff
@@ -1367,154 +1519,31 @@ func (cr *chunkRestore) restore(
...
-		encodeCompleted bool
-		totalKVs        []kvenc.KvPair
-		localChecksum   verify.KVChecksum
-		chunkOffset     int64
-		chunkRowID      int64
-	}
-	block.cond = sync.NewCond(new(sync.Mutex))
-	deliverCompleteCh := make(chan error, 1)
+	kvsCh := make(chan deliveredKVs, maxKVQueueSize)
+	deliverCompleteCh := make(chan deliverResult)
```
lonng: This channel should be closed properly, I think after L1450 is well.

lightning/restore/restore.go
GregoryIan: we should check it while we need send data into the input channel

lightning/kv/session.go
GregoryIan: one question: if we want to reduce memory allocation, is it the best way to put `kvPair` into memory pool?

lightning/restore/checkpoints.go
```diff
@@ -536,9 +542,10 @@ func (cpdb *MySQLCheckpointsDB) InsertEngineCheckpoints(ctx context.Context, tab
 				return errors.Trace(err)
 			}
 			for _, value := range engine.Chunks {
+				colPerm, _ := json.Marshal(value.ColumnPermutation)
 				_, err = chunkStmt.ExecContext(
 					c, tableName, engineID,
-					value.Key.Path, value.Key.Offset, value.Columns, value.ShouldIncludeRowID,
+					value.Key.Path, value.Key.Offset, colPerm,
```
GregoryIan: I only know we call `InsertEngineCheckpoints` after calling `populateChunks`.  If so, `columns` column always empty

lightning/restore/restore.go
```diff
@@ -1359,6 +1379,138 @@ func splitIntoDeliveryStreams(totalKVs []kvenc.KvPair, splitSize int) [][]kvenc.
...
+		}
+	}
+
+	return
+}
+
+func (cr *chunkRestore) saveCheckpoint(t *TableRestore, engineID int32, rc *RestoreController) {
+	rc.saveCpCh <- saveCp{
+		tableName: t.tableName,
+		merger: &RebaseCheckpointMerger{
```
GregoryIan: is it enough to only save it in L529?

lightning/restore/restore_test.go
```diff
@@ -26,30 +31,35 @@ func (s *restoreSuite) TestNewTableRestore(c *C) {
 	testCases := []struct {
 		name       string
 		createStmt string
-		errRegexp  string
 	}{
-		{"t1", "CREATE TABLE `t1` (`c1` varchar(5) NOT NULL)", ""},
-		{"t2", "CREATE TABLE `t2` (`c1` varchar(30000) NOT NULL)", "failed to ExecDDLSQL `mockdb`.`t2`:.*"},
-		{"t3", "CREATE TABLE `t3-a` (`c1-a` varchar(5) NOT NULL)", ""},
+		{"t1", "CREATE TABLE `t1` (`c1` varchar(5) NOT NULL)"},
+		// {"t2", "CREATE TABLE `t2` (`c1` varchar(30000) NOT NULL)"}, // no longer able to create this kind of table.
```
GregoryIan: is this case meaningless? or add some error cases?

tests/checkpoint_chunks/run.sh
```diff
@@ -33,7 +33,7 @@ done
 
 # Set the failpoint to kill the lightning instance as soon as one chunk is imported
```
GregoryIan: fix comment

lightning/mydump/csv_parser.go
```diff
@@ -126,24 +136,14 @@ func (parser *CSVParser) ReadRow() error {
...
 
 		case csvTokField:
-			parser.appendEmptyValues(emptySepCount)
-			emptySepCount = 0
-			parser.appendField(content, false, '\'')
-
-		case csvTokQuotedField:
-			parser.appendEmptyValues(emptySepCount)
+			parser.appendEmptyValues(emptySepCount - 1)
 			emptySepCount = 0
```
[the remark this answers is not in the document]
kennytm: @GregoryIan What do you mean?

lightning/kv/session.go
kennytm: This is a much bigger change than that record cache, and since the KV pairs are shared across threads it may be harder to ensure thread safety. I'd like to hold this off unless the memory profile shows otherwise.

lonng: Actually, I think it is better to maintain a free list in session struct.

lonng: @kennytm Yes, it's OK for lightning.

lightning/restore/restore.go
```diff
@@ -1367,260 +1520,103 @@ func (cr *chunkRestore) restore(
...
-		block.cond.Signal()
-		block.cond.L.Unlock()
 	}
 
-	block.cond.L.Lock()
-	block.encodeCompleted = true
-	block.cond.Signal()
-	block.cond.L.Unlock()
+	lastOffset, lastRowID := cr.parser.Pos()
+	kvsCh <- deliveredKVs{
```
GregoryIan: add `select` for it? like L1589

lightning/mydump/parser.go
```diff
@@ -77,22 +77,46 @@ type Chunk struct {
...
+
+const (
+	backslashEscapeFlavorNone backslashEscapeFlavor = iota
+	backslashEscapeFlavorMySQL
+	backslashEscapeFlavorMySQLWithNull
+
+	// LegacyNullSequence is a backward-compatibility internal representation
+	// of the escape sequence "\N". It is generated randomly to minimize chance
+	// of collision with user data. Used for CSV parsing only.
+	legacyNullSequence = "\xeaGs\xf5\x15\xb0\xa2%\x16\xf7z\x9d\xbe\xd8\xb3\x85"
```
GregoryIan: Any more explanation?

lightning/mydump/parser.go
```diff
@@ -160,6 +193,57 @@ func (parser *blockParser) readBlock() error {
...
 
+var unescapeRegexp = regexp.MustCompile(`(?s)\\.`)
+
+func unescape(
+	input string,
+	delim string,
+	escFlavor backslashEscapeFlavor,
+) string {
+	if len(delim) > 0 {
+		delim2 := delim + delim
```
GregoryIan: do need to verify `Delimiter` inputed by user if some guys input `\`?

lightning/mydump/parser.go
```diff
@@ -160,6 +193,57 @@ func (parser *blockParser) readBlock() error {
...
+			case 'n':
+				return "\n"
+			case 'r':
+				return "\r"
+			case 't':
+				return "\t"
+			case 'Z':
+				return "\x26"
+			default:
+				return substr[1:]
```
GregoryIan: is it ok to discard '\\'？

lightning/mydump/parser.go
```diff
@@ -175,44 +259,82 @@ func (parser *ChunkParser) ReadRow() error {
...
+		// the state after reading "VALUES"
+		stateValues
+
+		// the state while reading row values
+		stateRow
 	)
 
 	row := &parser.lastRow
-	st := stateRow
+	st := stateValues
```
GregoryIan: why not initial to `stateTableName`?

lightning/mydump/parser_test.go
```diff
@@ -19,11 +19,12 @@ import (
 	"strings"
 
 	. "github.com/pingcap/check"
+	"github.com/pingcap/errors"
```
GregoryIan: we don't add more test for it, only old cases

lightning/mydump/csv_parser.go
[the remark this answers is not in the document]
GregoryIan: no problem after I take a careful look

GregoryIan:
Rest LGTM

lightning/restore/restore.go
kennytm: Unfortunately no, the AllocBase needs to be larger than all content of `_tidb_rowid` (or the integer primary key), which value cannot be determined until we've read all data.

Added a comment for this.

lightning/restore/checkpoints.go
kennytm: Removed this and the checksum (always 0 too).

lightning/restore/restore_test.go
kennytm: In this PR we no longer parses the `CREATE TABLE` DDL, and instead directly unmarshal the JSON result from TiDB (calling `tables.TableFromMeta`). So yeah this case becomes meaningless since either you can't create a `VARCHAR(30000)` in TiDB, or you can and produced a proper JSON which Lightning accepts without error.

Anyway, added a separate unit test to ensure malformed table info will produce an error.

lightning/mydump/parser.go
kennytm: Never mind, this is not needed.

lightning/mydump/parser.go
kennytm: `\\` is would be unescaped as `\` just like all other normal characters.

```
mysql> select '\\\x\y\1';
+------+
| \xy1 |
+------+
| \xy1 |
+------+
1 row in set (0.00 sec)
```

lightning/mydump/parser.go
kennytm: Because an INSERT statement may have multiple rows, and this method only read one.

```sql
INSERT INTO `table` (`a`, `b`) VALUES (1, 2)   -- first ReadRow() call parse until here
,(3, 4)                                        -- second ReadRow() call parse until here
;
```

lightning/mydump/parser_test.go
kennytm: The old unit test didn't compile since the interface is changed. There's no new tests in this file.

tests/sqlmode/run.sh
```diff
@@ -0,0 +1,42 @@
+#!/bin/sh
```
lonng: Please add the license header.

lonng:
Rest LGTM

kennytm:
@GregoryIan PTAL again

lightning/mydump/parser.go
GregoryIan: ok

tests/checkpoint_chunks/run.sh
GregoryIan: in fact it fails while kvs > xxx

kennytm: How about this? 

```sh
# Set the failpoint to kill the lightning instance as soon as 
# one file (after writing totally $ROW_COUNT rows) is imported
```

Since one chunk contains `$ROW_COUNT`, the effect is the same. The `>=` ensures we don't accidentally break in the middle of the file which increases the retry count.

GregoryIan: ok

GregoryIan:
LGTM

PR #150, opened 2019-03-20
restore: eager release index engine worker after imported

lonng:

Index engine worker will be released after `TableRestore.postProcess` in the current implementation, and `TableRestore.postProcess` contains `checksum` and `analyze` steps which should not keep index worker.

Extract the table restore engines logic to an individual method, and eager release index engine worker after all data engines and index engine were been imported.

### Check List 

Tests 

 - Integration test

Side effects

N/A

Related changes

N/A


GregoryIan:
LGTM

PR #151, opened 2019-03-21
checkpoint: fix checkpoint not updated in some scenario

amyangfei:

Test `checkpoint_error_destroy` in integration got error randomly, both in local environment and JenkinsCI.
Lightning could exit before all checkpoints are saved, as the `WaitGroup` is not always 
added before this line https://github.com/pingcap/tidb-lightning/blob/master/lightning/restore/restore.go#L204 returns

1. After `RestoreController.Run` is returned, no more messages will be sent to `rc.saveCpCh`, so we can close it safely and ensure all `saveCp` are consumed
2. remove some unused code

### Check List 

Tests 

 - Integration test

### DIFF
--- a/lightning/common/util.go
+++ b/lightning/common/util.go
@@ -39,10 +39,6 @@ const (
 	defaultMaxRetry = 3
 )
 
-func Percent(a int, b int) string {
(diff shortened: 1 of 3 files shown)


amyangfei:
PTAL @kennytm @lonng

PR #154, opened 2019-03-27
 kv,restore: log which value caused conversion failure

kennytm:

1. When there is invalid data, the current log is not able to reveal what is the cause.
2. When a column is missing, Lightning will fill in default values for these columns, but this behavior may not be obvious.

1. When KV encode failed, log the file path, row content, column index, name and type in additional to the error.
2. When a column is missing, log a warning about the missing column.

### Check List 

Tests 

 - Integration test

Side effects

Related changes


lightning/kv/sql2kv.go
```diff
@@ -47,11 +50,76 @@ func (kvcodec *TableKVEncoder) Close() {
...
+			enc.AddInt("col", index)
+			enc.AddString("kind", kindStr[kind])
+			enc.AddString("val", str)
+			return nil
+		}))
+	}
+	return nil
+}
+
+func logKVConvertFailed(logger log.Logger, row []types.Datum, j int, colInfo *model.ColumnInfo, err error) {
```
GregoryIan: Will not brush too much？

GregoryIan:
Rest LGTM

lightning/kv/sql2kv.go
kennytm: What do you mean?

GregoryIan: output too many logs

kennytm: It will print at most a few per table with errors, since returning error should stop importing such tables.

GregoryIan:
LGTM

Issue #131, opened 2019-02-13
Require less configuration


morgo:
## Feature Request

**Is your feature request related to a problem? Please describe:**

Currently lightning requires a lot of configuration:
* tikv-importer address
* mydumper data-source-dir
* tidb host
* pd host

I would like to find a way where it can be used with minimal configuration.  This helps improve convenience/support notice users and experiments.

**Describe the feature you'd like:**

I am open to ideas on implementation:
- Can lightning discover a tidb-host and tikv-importer host from PD?  That might not work perfectly, since it won't know which one is closest. I think it could absolutely discover the pd host from TiDB though (making the configuration of PD optional).
- Could lightning be embedded/bundled as a TiDB plugin?  it would be nice to load data directly with SQL-like syntax:
`LIGHTNING LOAD 's3://path/to/mydumper'` (using the local tidb and learning pd from it).
- Can tikv-importer somehow become embedded too?
- Could the mydumper data-source-dir just be moved from configuration, to argument `$1` for `tidb-lightning`?

**Describe alternatives you've considered:**

This is really only about improving convenience/usability for casual use cases, so there are many alternative implimentations.  There are a lot of users that don't want to edit configuration files, but rather just have a tool setup and running with no effort.

**Teachability, Documentation, Adoption, Optimization:**

kennytm:
> Can lightning discover a tidb-host and tikv-importer host from PD?  That might not work perfectly, since it won't know which one is closest. I think it could absolutely discover the pd host from TiDB though (making the configuration of PD optional).

Thanks, looks like we could find the PD address from http://tidb-ip:10080/settings

Importer isn't registered on PD though, so we can't use PD to discover Importer.

> Could the mydumper data-source-dir just be moved from configuration, to argument `$1` for `tidb-lightning`?

Yes

> Could lightning be embedded/bundled as a TiDB plugin? 
>
> Can tikv-importer somehow become embedded too?

cc @jackysp?

A plugin is basically a *.so library, we could place the Lightning/Importer code inside the plugin, or make the plugin a front-end which controls Lightning/Importer on another machine.

Placing the code directly inside the plugin is the same as the "mixed deployment" strategy which we no matter recommend. Lightning/Importer are resource intensive programs and doing so may bring down the cluster due to using up all CPU and network bandwidth (making the cluster unresponsive).

```
+---------------+
| TiDB          |
| +-----------+ |
| | Lightning | |
| +-----------+ |  +------+
[...]

PR #157, opened 2019-03-28
config: Allow overriding some config from command line

kennytm:

Addresses one part of #131, makes `config.toml` unnecessary.

Added the following flags:

* -L
* -d
* --log-file
* --tidb-host, --tidb-port, --tidb-user, --tidb-status
* --pd-urls
* --importer

Additionally, *removed* the following flags from tidb-lightning (use the corresponding flags from tidb-lightning-ctl instead):

* --compact
* --switch-mode

### Check List 

Tests 

 - Integration test

Side effects

    * unspecified `--config` now ~~also searches `conf/tidb-lightning.toml`, and~~ fallbacks to use the default config ~~if the file cannot be found~~.
    * `--compact` and `--switch-mode` are removed.
    * **This PR should not be cherry-picked to 2.1**. Consider not merging it until 2.1.8 is frozen.

Related changes


lightning/config/config.go
```diff
@@ -176,38 +180,92 @@ func NewConfig() *Config {
...
 }
 
 func (cfg *Config) Load() error {
-	if cfg.printVersion {
-		fmt.Println(common.GetRawInfo())
-		return flag.ErrHelp
+	// search the default config path if none is specified.
+	if cfg.ConfigFile == "" {
+		for _, path := range defaultConfigPaths {
+			if _, err := os.Stat(path); err == nil {
```
GregoryIan: not good design, if user doesn't specify config file path, we should use default configuration values, its more nature

kennytm: The behavior is inherited from 2.x where we default to read form `tidb-lightning.toml` if `-config` is not specified. We could break this behavior?

PR #158, opened 2019-04-03
tests: add kill lightning test in checkpoint_chunks case

amyangfei:
add test case to cover lightning is killed durning data import

use gofail to control lightning to kill itself after one chunk is imported

### Check List 

Tests 

 - Integration test

### DIFF
--- a/tests/checkpoint_chunks/run.sh
+++ b/tests/checkpoint_chunks/run.sh
@@ -20,7 +20,7 @@ DBPATH="$TEST_DIR/cpch.mydump"
 CHUNK_COUNT=5
 ROW_COUNT=1000
 
-verify_checkpoint_op() {
+verify_checkpoint_noop() {
     # After everything is done, there should be no longer new calls to WriteEngine/CloseAndRecv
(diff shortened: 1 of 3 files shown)


amyangfei:
PTAL @kennytm @lonng

tests/checkpoint_chunks/run.sh
```diff
@@ -47,21 +59,43 @@ for i in $(seq "$CHUNK_COUNT"); do
...
+# If checkpoint does work, this should only kill $CHUNK_COUNT instances of lightnings.
+export GOFAIL_FAILPOINTS='github.com/pingcap/tidb-lightning/lightning/restore/KillIfImportedChunk=return'
+
+for i in $(seq "$CHUNK_COUNT"); do
+    echo "******** Importing Chunk Now (step $i/$CHUNK_COUNT) ********"
+    run_lightning
+done
+
+set +e
+while true; do
```
kennytm: Add a retry limit to avoid getting stuck in infinite loop.

PR #156, opened 2019-03-28
restore: give priority to small tables for importing

lonng:
### What problem does this PR solve? 

Put the large table in the front of the slice which can avoid large table take a long time to import and block small table to release index worker.

### What is changed and how it works?

Sorting tables before restore.

### Check List 

Tests 

 - Unit test
 - Integration test

Side effects

N/A

Related changes

N/A

PR #161, opened 2019-04-09
Fix interpretation of integers in a BIT column

kennytm:

Allows the following to pass:

```sql
CREATE TABLE t(a BIT(1));
INSERT INTO t VALUES (1);
```

Previously all integers are written as a string datum, and we use `CastValue` to convert the string into the target types on demand, the result is the same for numbers. But turns out casting a string `'1'` and integer `1` into BIT type behave differently, the former will become the ASCII value instead (i.e. `x'31'` instead of `x'01'`). 

So in this PR we just add a special token for integers from the parser, and convert them into integer datum early on.

This PR does not change handling of CSV files, which always treat all fields as strings.

### Check List 

Tests 

 - Unit test
 - Integration test

Side effects

Related changes

    * Needs to be cherry-picked before 2.1.8 release to avoid regression.


kennytm:
PTAL @lonng @GregoryIan

lightning/mydump/loader_test.go
```diff
@@ -62,8 +62,8 @@ func (s *testMydumpLoaderSuite) TestLoader(c *C) {
 	}{
 		{name: "i", dataFiles: 1},
 		{name: "report_case_high_risk", dataFiles: 1},
-		{name: "tbl_autoid", dataFiles: 1},
 		{name: "tbl_multi_index", dataFiles: 1},
+		{name: "tbl_autoid", dataFiles: 1},
```
lonng: It'OK, but seems that is an unnecessary change, what does it affect?

kennytm: Oh the test was broken by #156, seems `tbl_multi_index` is smaller than `tbl_autoid`.

lightning/mydump/parser.go
```diff
@@ -154,12 +155,51 @@ const (
...
 	tokSingleQuoted
 	tokDoubleQuoted
 	tokBackQuoted
 	tokUnquoted
 )
 
+// String implements the fmt.Stringer interface
+//
+// Mainly used for debugging a token.
+func (t token) String() string {
```
lonng: I think the better way to implement the `fmt.Stringer` here is using a string array because of the token enum is an autoincrement number. For example,
```go
var tokStringify = [...] {
	tokNil: "<Nil>",
	tokRowBegin: "RowBegin",
	// ...
}
func (t token) String() string {
	if t < tokNil || t > tokUnquoted || tokStringify[t] == "" {
		return fmt.Sprintf("unknown(%d), t)
	}
	return tokStringify[t]
}
```

lightning/mydump/loader_test.go
lonng: Got it.

GregoryIan:
should we also add test cases for `enum` or `set` type?

kennytm:
PTAL @GregoryIan 411c4766c97f0fe30cc2945c2b70a3036e19b81b

Also tried to add a test for hex literals, but turns out TiDB and MySQL are incompatible again (pingcap/tidb#10154) 🤷

GregoryIan:
🤣 LGTM

PR #162, opened 2019-04-11
*: use pingcap/log (zap) for logging 

kennytm:

Switch to use the [uniform log format](https://github.com/tikv/rfcs/blob/master/text/2018-12-19-unified-log-format.md). 

The log library is changed from `logrus` to `zap`, a structured logger.

Due to how structured logging usually works, several interfaces are rewritten.

* *Child loggers* are heavily used to abstract away contextual information in the log, e.g. the table name, engine UUID etc. Many types which originally compute these tagging information on-line are changed to storing a child logger instead.
* The 3 functions `QueryRowWithRetry`/`TransactWithRetry`/`ExecWithRetry` originally writes log without any contextual information. These are modified to accept a logger. The common arguments are abstracted into an `SQLWithRetry` type.

All logging infrastructures (e.g. `common.AppLogger`) are moved from `common` into a new `log` package. 

* Like `common.AppLogger`, we need to define our own logger `log.L()` distinct from the true global one `"github.com/pingcap/log".L()` so we could control TiDB's logging level independently (though perhaps we could investigate whether merging the two is possible nowadays). 
* Stack traces of error level logs are entirely disabled, and stack traces of errors are by default inhibited by the `log.ShortError` function, to avoid flooding the log with noise.
* A common pattern used in Lightning is:
     1. Log that a task has started
     2. Perform a task
     3. Log that a task has ended, and display the total time taken.  
  This is codified into the `log.Task` type.

Several redundant error logs are eliminated (e.g. if `f()` calls `g()` calls `h()`, and `h()` logs an error, then `g()` and `f()` will not log the error again). 

Following pingcap/tidb#10025's lead, when refactoring a function, if an error is known to be a custom error, `errors.Trace` will no longer be applied. This will be done more thoroughly in the future, after unit-test coverage is improved.

### Check List 

Tests 

 - Unit test
 - Integration test

Side effects

    * Log format is modified

Related changes


lightning/common/util.go
```diff
@@ -182,35 +182,17 @@ func IsRetryableError(err error) bool {
...
+// affected by whether the error is canceling or not.
 //
 // This function returns `false` (not a context-canceled error) if `err == nil`.
 func IsContextCanceledError(err error) bool {
-	err = errors.Cause(err)
-	return err == context.Canceled || status.Code(err) == codes.Canceled
+	return log.IsContextCanceledError(err)
+}
+
+func ShouldLogError(err error) bool {
```
kennytm: (I'll remove this function later.)

kennytm:
Output looks like:

Screenshot 2019-04-17 at 02 53 56

lightning/common/once_error.go
```diff
@@ -28,16 +28,13 @@ type OnceError struct {
...
 // If this method is called multiple times, only the first call is effective.
-func (oe *OnceError) Set(tag string, e error) {
+func (oe *OnceError) Set(e error) {
 	if e != nil {
 		oe.lock.Lock()
 		if oe.err == nil {
 			oe.err = e
 		}
 		oe.lock.Unlock()
-		if ShouldLogError(e) {
```
csuzhangxc: remove `ShouldLogError` in _common/util.go_?

lightning/kv/importer.go
```diff
@@ -102,43 +103,34 @@ func (importer *Importer) SwitchMode(ctx context.Context, mode sst.SwitchMode) e
 			Mode: mode,
 		},
 	}
-	timer := time.Now()
 
+	task := log.With(zap.Stringer("mode", mode)).Begin(zap.DebugLevel, "switch mode")
```
csuzhangxc: why do we use `DebugLevel`?

kennytm: Originally there was no "switch mode start" log, so this was set to debug level to keep the log from showing by default.

lightning/restore/restore.go
```diff
@@ -1289,38 +1271,26 @@ type RemoteChecksum struct {
...
 		return nil, errors.Trace(err)
 	}
 	// set it back finally
-	defer func() {
-		err = UpdateGCLifeTime(ctx, db, ori)
-		if err != nil && common.ShouldLogError(err) {
-			common.AppLogger.Errorf("[%s] update tikv_gc_life_time error %v", table, errors.ErrorStack(err))
-		}
-	}()
+	defer UpdateGCLifeTime(ctx, db, ori)
```
csuzhangxc: log the returned error is better? so we can set it back manually if needed.

lightning/log/log.go
```diff
@@ -0,0 +1,190 @@
...
+	case err == nil:
+		level = task.level
+		verb = " completed"
+	case IsContextCanceledError(err):
+		level = zap.DebugLevel
+		verb = " canceled"
+		extraFields = nil
+	default:
+		verb = " failed"
+		extraFields = nil
```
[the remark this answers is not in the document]
kennytm: Yes. Those extra information may be incorrect if the task failed. Added comment.

lightning/log/log.go
```diff
@@ -0,0 +1,190 @@
...
+// level.
+func SetLevel(level zapcore.Level) zapcore.Level {
+	oldLevel := appLevel.Level()
+	appLevel.SetLevel(level)
+	return oldLevel
+}
+
+// ShortError contructs a field which only records the error message without the
+// verbose text (i.e. excludes the stack trace).
+func ShortError(err error) zap.Field {
```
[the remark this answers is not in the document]
kennytm: For Lightning, almost always `ShortError`. No advice for other projects. Will add a comment.

lightning/restore/restore.go
```diff
@@ -481,19 +497,12 @@ func (rc *RestoreController) restoreTables(ctx context.Context) error {
 	for _, dbMeta := range rc.dbMetas {
 		dbInfo, ok := rc.dbInfos[dbMeta.Name]
 		if !ok {
-			common.AppLogger.Errorf("database %s not found in rc.dbInfos", dbMeta.Name)
-			continue
+			return errors.Errorf("database %s not found in rc.dbInfos", dbMeta.Name)
```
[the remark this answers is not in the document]
kennytm: Yes unify behavior with L505.

lightning/restore/restore.go
```diff
@@ -817,11 +833,7 @@ func (t *TableRestore) importEngine(
 	if rc.cfg.PostRestore.Level1Compact &&
 		atomic.CompareAndSwapInt32(&rc.compactState, compactStateIdle, compactStateDoing) {
 		go func() {
-			err := rc.doCompact(ctx, Level1Compact)
-			if err != nil {
-				// log it and continue
-				common.AppLogger.Warnf("compact %d failed %v", Level1Compact, err)
-			}
+			rc.doCompact(ctx, Level1Compact)
```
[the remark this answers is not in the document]
kennytm: Yes. Added comment.

lightning/restore/restore.go
```diff
@@ -920,9 +925,7 @@ func (rc *RestoreController) switchToNormalMode(ctx context.Context) error {
 }
 
 func (rc *RestoreController) switchTiKVMode(ctx context.Context, mode sstpb.SwitchMode) {
-	if err := rc.importer.SwitchMode(ctx, mode); err != nil {
-		common.AppLogger.Warnf("cannot switch to %s mode: %v", mode.String(), err)
-	}
+	rc.importer.SwitchMode(ctx, mode)
```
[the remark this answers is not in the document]
kennytm: ditto

GregoryIan:
@july2993 @lonng PTAL

lightning/verification/checksum.go
```diff
@@ -88,3 +89,10 @@ func (c *KVChecksum) SumSize() uint64 {
 func (c *KVChecksum) SumKVS() uint64 {
 	return c.kvs
 }
+
+func (c *KVChecksum) MarshalLogObject(encoder zapcore.ObjectEncoder) error {
```
july2993: add comment for this? 
ipml zapcore#ObjectMarshaler

lightning/restore/restore.go
```diff
@@ -96,13 +99,17 @@ type errorSummaries struct {
 func (es *errorSummaries) emitLog() {
 	es.Lock()
 	defer es.Unlock()
+
+	logger := log.L()
```
lonng: It's OK but better to move this call to if statement.

lightning/restore/restore.go
```diff
@@ -1209,31 +1215,24 @@ func (t *TableRestore) initializeColumns(columns []string, ccp *ChunkCheckpoint)
...
 func (tr *TableRestore) restoreTableMeta(ctx context.Context, db *sql.DB) error {
-	timer := time.Now()
-
-	err := AlterAutoIncrement(ctx, db, tr.tableMeta.DB, tr.tableMeta.Name, tr.alloc.Base()+1)
-	if err != nil {
-		return errors.Trace(err)
-	}
-	common.AppLogger.Infof("[%s] alter table set auto_id takes %v", common.UniqueTable(tr.tableMeta.DB, tr.tableMeta.Name), time.Since(timer))
-	return nil
+	return AlterAutoIncrement(ctx, db, tr.tableName, tr.alloc.Base()+1)
```
lonng: I think we can remove this function if only call another function directly.

PR #165, opened 2019-04-16
*: replace gofail with new failpoint implementation

lonng:
Signed-off-by: Lonng <chris@lonng.org>

Replace https://github.com/pingcap/gofail with the new failpoint implementation https://github.com/pingcap/failpoint.

1. Replace all failpoint injections
2. Replace related integration tests

### Check List 

Tests 

 - Unit test
 - Integration test

Side effects

N/A

Related changes


Makefile
```diff
@@ -107,8 +105,10 @@ update:
 	GO111MODULE=on go mod verify
 	GO111MODULE=on go mod tidy
 
-gofail-enable:
-	$(GOFAIL_ENABLE)
+failpoint-enable:
+	@hash failpoint-ctl || $(GO) get -v github.com/pingcap/failpoint/failpoint-ctl
```
kennytm: Let's create a Makefile target so we don't need to repeat this 3 times

```makefile
install-failpoint:
 	@hash failpoint-ctl || $(GO) get -v github.com/pingcap/failpoint/failpoint-ctl

failpoint-enable: install-failpoint
 	$(FAILPOINT_ENABLE)
```

lightning/restore/restore.go
```diff
@@ -357,43 +358,37 @@ func (rc *RestoreController) listenCheckpointUpdates() {
...
-
-		// gofail: var FailIfIndexEngineImported int
-		// if merger, ok := scp.merger.(*StatusCheckpointMerger); ok && merger.EngineID == wholeTableEngineID && merger.Status == CheckpointStatusIndexImported && FailIfIndexEngineImported > 0 {
-		// 	rc.checkpointsWg.Done()
-		// 	rc.checkpointsWg.Wait()
-		// 	panic("forcing failure due to FailIfIndexEngineImported")
-		// }
-		// goto RETURN3
+		failpoint.Inject("FailIfImportedChunk", func(val failpoint.Value) {
+			if merger, ok := scp.merger.(*ChunkCheckpointMerger); ok && merger.Checksum.SumKVS() >= uint64(val.(int)) {
```
kennytm: 🤔.

I wonder if there's plan to support

```go
failpoint.Inject("FailIfImportedChunk", func(val uint64) {
```

to avoid such awkward cast.

lightning/restore/restore.go
```diff
@@ -1239,7 +1235,8 @@ func (tr *TableRestore) importKV(ctx context.Context, closedEngine *kv.ClosedEng
 	metric.ImportSecondsHistogram.Observe(dur.Seconds())
 	common.AppLogger.Infof("[%s] kv deliver all flushed, takes %v", closedEngine.Tag(), dur)
 
-	// gofail: var SlowDownImport struct{}
+	// TODO: implement failpoint inject empty
+	failpoint.Inject("SlowDownImport", func() {})
```
kennytm: For now let's make it 

```go
failpoint.Inject("SlowDownImport", func(val failpoint.Value) {
    time.Sleep(val.(int) * time.Millisecond)
})
```

to keep it working. (If we could implement the `sleep` thing before this PR is merged then disregard this comment.)

Makefile
```diff
@@ -83,7 +81,7 @@ lightning_for_integration_test:
 		-coverpkg=github.com/pingcap/tidb-lightning/... \
```
GregoryIan: what if it exists with code != 0

lightning/restore/restore.go
lonng: It's hard to implement it. 😞

lightning/restore/restore.go
lonng: I will fix it to tomorrow

Makefile
lonng: It workspace to be dirty.

GregoryIan: any way to fix it?

lonng: I will fix it.

GregoryIan:
LGTM

PR #168, opened 2019-04-22
main: sync log before exit + update failpoint dep

kennytm:

We need to sync the logger before exit. https://godoc.org/go.uber.org/zap#Logger.Sync

Added a call to `Sync()`, modeled after TiDB https://github.com/pingcap/tidb/blob/b864fac504f9f059a39a7d477e923440f6d88a55/tidb-server/main.go#L176-L182.

Also, updated failpoint deps, see pingcap/dm#122.

### Check List 

Tests 

    * `make` works

Side effects

Related changes

### DIFF
--- a/cmd/tidb-lightning/main.go
+++ b/cmd/tidb-lightning/main.go
@@ -64,10 +64,19 @@ func main() {
 	}()
 
 	err = app.Run()
+	logger := log.L()
 	if err != nil {
-		log.L().Error("tidb lightning encountered error", zap.Error(err))
-		os.Exit(1)
+		logger.Error("tidb lightning encountered error", zap.Error(err))
+	} else {
+		logger.Info("tidb lightning exit")
 	}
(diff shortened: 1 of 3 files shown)


kennytm:
PTAL @july2993 @lonng

csuzhangxc:
`Sync()` will also flush child logger's buffered entries, right?

kennytm:
@csuzhangxc Yes. 

Calling [`With`](https://github.com/uber-go/zap/blob/badef736563f8135e0add3232b06f58656da03c8/zapcore/core.go#L72-L76) on an I/O Core would create a "clone" which [share the same WriteSyncer](https://github.com/uber-go/zap/blob/badef736563f8135e0add3232b06f58656da03c8/zapcore/core.go#L111). This means `parent.Sync()` and `child.Sync()` are both equivalent.

PR #170, opened 2019-04-23
kv: fix handling of column default values

kennytm:

Properly fill in the default value for a missing column.

* if the column is AUTO_INCREMENT, fill in with row_id (assume it is missing for the entire table instead of just a few values)
* if the column has DEFAULT, fill in that value
* (if the column is NULL, fill in NULL.)

### Check List 

Tests 

 - Integration test

Side effects

Related changes


GregoryIan:
LGTM

lightning/kv/sql2kv.go
```diff
@@ -69,13 +69,16 @@ func (kvcodec *TableKVEncoder) Encode(
...
 			value, err = table.CastValue(kvcodec.se, row[j], col.ToInfo())
 			if err == nil {
 				value, err = col.HandleBadNull(value, kvcodec.se.vars.StmtCtx)
 			}
+		} else if mysql.HasAutoIncrementFlag(col.Flag) {
+			// we still need a conversion, e.g. to catch overflow with a TINYINT column.
+			value, err = table.CastValue(kvcodec.se, types.NewIntDatum(rowID), col.ToInfo())
 		} else {
-			value, err = table.GetColOriginDefaultValue(kvcodec.se, col.ToInfo())
+			value, err = table.GetColDefaultValue(kvcodec.se, col.ToInfo())
```
csuzhangxc: not handle some cases like `current_timestamp` anymore?

kennytm: `CURRENT_TIMESTAMP` does work, it currently fetches the actual current time 🤔. (Before this PR it was always NULL.) Gonna change to a fixed timestamp stored in the checkpoint to ensure determinism in a future PR.

PR #172, opened 2019-04-25
tests,restore: re-enable the exotic_filenames test

kennytm:

The `exotic_filenames` test was disabled due to pingcap/tidb#8259. Since pingcap/tidb#8302 has been merged quite long ago, we could re-enable this test case.

The test suggests there's a bug in handling database names with `` ` `` in it, so we also fixed the problem in this PR.

Added one more test to check non-ASCII database name (the test for non-ASCII *table* name already existed before).

### Check List 

Tests 

 - Integration test

Side effects

Related changes


kennytm:
PTAL @csuzhangxc @lonng

PR #173, opened 2019-04-25
config: automatically discover tidb.pd-addr and tidb.port if not provided

kennytm:

Part of #131. 

The setting `tidb.port` and `tidb.pd-addr` can be automatically discovered through the status API, allowing these to be omitted while Lightning could still obtain the correct values.

Change the default values of `tidb.port` from 4000 to 0 and `tidb.pd-addr` from "127.0.0.1:2379" to "".

If these values are empty, reach for `http://{tidb.host}:{tidb.status-port}/settings`, and use these settings instead.

### Check List 

Tests 

 - Unit test
 - Integration test

Side effects

 - Increased code complexity

Related changes


lightning/config/config.go
```diff
@@ -286,9 +290,47 @@ func (cfg *Config) Load() error {
...
+				}
+			}
+			if cfg.TiDB.Port <= 0 {
+				cfg.TiDB.Port = settings.Port
+			}
+		}
+		if len(cfg.TiDB.PdAddr) == 0 {
+			pdAddrs := strings.Split(settings.Path, ",")
+			cfg.TiDB.PdAddr = pdAddrs[0] // FIXME support multiple PDs once importer can.
+		}
```
GregoryIan: check `cfg.TiDB.Port` and `cfg.TiDB.PdAddr` again, if they are empty, throw an error

lightning/config/config.go
```diff
@@ -286,9 +290,47 @@ func (cfg *Config) Load() error {
...
+	if cfg.TiDB.Port <= 0 || len(cfg.TiDB.PdAddr) == 0 {
+		resp, err := http.Get(fmt.Sprintf("http://%s:%d/settings", cfg.TiDB.Host, cfg.TiDB.StatusPort))
+		if err != nil {
+			return errors.Annotate(err, "cannot fetch settings from TiDB, please manually fill in `tidb.port` and `tidb.pd-addr`")
+		}
+		defer resp.Body.Close()
+		if resp.StatusCode != http.StatusOK {
+			return errors.Errorf("TiDB settings returned %s, please manually fill in `tidb.port` and `tidb.pd-addr`", resp.Status)
+		}
+		var settings struct {
```
GregoryIan: use `tidb/config.Config` directly, even it contains too many other fields whose we don't need

lightning/config/config.go
```diff
@@ -286,9 +290,47 @@ func (cfg *Config) Load() error {
...
+		}
+		if cfg.TiDB.Port <= 0 {
+			_, portStr, err := net.SplitHostPort(settings.AdvertiseAddress)
+			if err == nil {
+				port, err := strconv.Atoi(portStr)
+				if err == nil {
+					cfg.TiDB.Port = port
+				}
+			}
+			if cfg.TiDB.Port <= 0 {
```
GregoryIan: put it before L317?

GregoryIan:
Rest LGTM

lightning/config/config.go
kennytm: @GregoryIan but `cfg.TiDB.Port` may be modified by line 321

GregoryIan: in fact I mean priority of `settings.Port` is higher than `settings.AdvertiseAddress`

kennytm: If `settings.AdvertiseAddress` has provided a valid port, the port would be set to that and `settings.Port` is ignored as `cfg.TiDB.Port > 0`. So instead `settings.AdvertiseAddress` has higher priority than `settings.Port`. This has been checked in the unit test.

GregoryIan: I ask tidb guys, they told me to use `Port`, it seems `AdvertiseAddress` just is used for PD 😖

kennytm: So it is not supported to put a port-forwarder  in front of TiDB?

GregoryIan: According to the information I understand, it is indeed

GregoryIan:
LGTM

lonng:
@kennytm Please resolve the conflicts

kennytm:
PTAL @lonng

PR #175, opened 2019-04-27
config: reduce default table-concurrency from 8 to 6

kennytm:

The current default table-concurrency is 8, meaning up to 8 data engines will be opened. Then the default index-concurrency is 2, so together up to 10 engines will be opened. But Importer's default max-open-engine is 8, so with the default settings we will hit the TooManyOpenEngines error.

Reduce the default table-concurrency to 6.

### Check List 

Tests 

 - No code

Side effects

Related changes


GregoryIan:
LGTM

kennytm:
PTAL @lonng

PR #193, opened 2019-05-21
Add unit tests for 'restore.go' (TableRestore and chunkRestore)

kennytm:

Improve unit test coverage of restore package, part 2 of N.

Add unit tests (29% → 40%).

Performed some function extraction to allow for more focused testing. 

### Check List 

Tests 

 - Unit test

Side effects

Related changes

### DIFF
--- a/lightning/restore/restore.go
+++ b/lightning/restore/restore.go
@@ -1642,10 +1642,10 @@ func (cr *chunkRestore) restore(
 		kvEncoder.Close()
 		kvEncoder = nil
 		close(kvsCh)
-		close(deliverCompleteCh)
 	}()
 
 	go func() {
+		defer close(deliverCompleteCh)
 		dur, err := cr.deliverLoop(ctx, kvsCh, t, engineID, dataEngine, indexEngine, rc)
 		deliverCompleteCh <- deliverResult{dur, err}
 	}()
--- a/lightning/restore/restore.go
+++ b/lightning/restore/restore.go
@@ -1635,20 +1635,20 @@ func (cr *chunkRestore) restore(
 		t.encTable,
 		rc.cfg.TiDB.SQLMode,
 	)
+	kvsCh := make(chan deliveredKVs, maxKVQueueSize)
+	deliverCompleteCh := make(chan deliverResult)
+
 	defer func() {
 		kvEncoder.Close()
 		kvEncoder = nil
+		close(kvsCh)
+		close(deliverCompleteCh)
 	}()
-
-	kvsCh := make(chan deliveredKVs, maxKVQueueSize)
-	deliverCompleteCh := make(chan deliverResult)
 
 	go func() {
 		dur, err := cr.deliverLoop(ctx, kvsCh, t, engineID, dataEngine, indexEngine, rc)
 		deliverCompleteCh <- deliverResult{dur, err}
 	}()
-
-	defer close(kvsCh)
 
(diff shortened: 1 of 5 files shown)


kennytm:
PTAL @lonng @july2993

lightning/restore/restore.go
```diff
@@ -1607,42 +1600,65 @@ outside:
...
-		break
-	case <-ctx.Done():
-		return ctx.Err()
-	case deliverResult := <-deliverCompleteCh:
-		if deliverResult.err == nil {
-			panic("unexpected: deliverCompleteCh prematurely fulfilled with no error")
-		}
-		return errors.Trace(deliverResult.err)
+	err = send(deliveredKVs{kvs: nil, offset: lastOffset, rowID: lastRowID})
+	return
```
lonng: I think it's better to move L1621-L1622 to above L1590 is better and replace `break` with `return`

lightning/restore/restore.go
```diff
@@ -1607,42 +1600,65 @@ outside:
...
+
+	kvsCh := make(chan deliveredKVs, maxKVQueueSize)
+	deliverCompleteCh := make(chan deliverResult)
+
+	go func() {
+		dur, err := cr.deliverLoop(ctx, kvsCh, t, engineID, dataEngine, indexEngine, rc)
+		deliverCompleteCh <- deliverResult{dur, err}
+	}()
+
+	defer close(kvsCh)
```
lonng: `deliverCompleteCh` should be closed also.

lightning/restore/restore_test.go
```diff
@@ -76,3 +96,689 @@ func (s *restoreSuite) TestNewTableRestoreFailure(c *C) {
...
+	fakeDataDir := c.MkDir()
+	fakeDataFilesCount := 6
+	fakeDataFilesContent := []byte("INSERT INTO `table` VALUES (1, 2, 3);")
+	c.Assert(len(fakeDataFilesContent), Equals, 37)
+	fakeDataFiles := make([]string, 0, fakeDataFilesCount)
+	for i := 1; i <= fakeDataFilesCount; i++ {
+		fakeDataPath := path.Join(fakeDataDir, fmt.Sprintf("db.table.%d.sql", i))
+		err = ioutil.WriteFile(fakeDataPath, fakeDataFilesContent, 0644)
+		c.Assert(err, IsNil)
+		fakeDataFiles = append(fakeDataFiles, fakeDataPath)
```
lonng: All files should be deleted in `TearDownTest`.

lightning/restore/restore.go
kennytm: No, line 1621-1623 will be hit by Line 1575 as well.

GregoryIan:
LGTM

lightning/restore/restore_test.go
kennytm: I'm going to move these into `SetUpSuite` since the data source is going to be immutable anyway.

kennytm: Also, `c.MkDir()` will return a [fresh directory](https://github.com/pingcap/check/blob/67f458068fc864dabf17e38d4d337f28430d13ed/check.go#L159) for every call, so the stale data from the previous tests won't affect the next ones anyway.

kennytm:
@lonng PTAL again. Moved the close inside the goroutine, otherwise the `<-` may crash due to execution order.

PR #198, opened 2019-06-05
config,lightning: Implements server mode 

kennytm:

As a prerequisite of the "web interface", we'd like to have Lightning keep running even after import is finished. 

The behavior overlaps with a feature request which turns Lightning into a server, which we could submit tasks to it (similar to DM-master). The conversion is straightforward, so why not 🙃.

1. Split config into two classes: "Global config" and "Task config" (both  are subset of `tidb-lightning.toml`, unlike DM task settings). 
    * Also refactored the existing argument parsing code to cut down the code duplication.

2. Added an HTTP endpoint which allows user to upload a Task config toml. 

3. Changed the public `Run()` method to loop for those uploaded tasks in server mode. Essentially like this:

    ```go
    func Run() {
        if !ServerMode {
            run(globalConfig)
        } else {
            for taskConfig := range uploadedTaskConfigs {
                run(taskConfig)
            }
        }
    }
    ```

### Check List 

Tests 

 - Unit test
 - Integration test

Side effects

 - Increased code complexity

Related changes

### DIFF
--- a/cmd/tidb-lightning/main.go
+++ b/cmd/tidb-lightning/main.go
@@ -14,14 +14,12 @@
 package main
 
 import (
-	"flag"
 	"fmt"
 	_ "net/http/pprof"
 	"os"
 	"os/signal"
 	"syscall"
 
-	"github.com/pingcap/errors"
 	"github.com/pingcap/tidb-lightning/lightning"
 	"github.com/pingcap/tidb-lightning/lightning/config"
 	"github.com/pingcap/tidb-lightning/lightning/log"
@@ -38,16 +36,7 @@ func setGlobalVars() {
 func main() {
 	setGlobalVars()
 
-	cfg, err := config.LoadConfig(os.Args[1:])
-	switch errors.Cause(err) {
-	case nil:
-	case flag.ErrHelp:
-		os.Exit(0)
-	default:
-		fmt.Println("Failed to parse command flags: ", err)
-		os.Exit(1)
-	}
-
+	cfg := config.Must(config.LoadGlobalConfig(os.Args[1:], nil))
 	app := lightning.New(cfg)
 
 	sc := make(chan os.Signal, 1)
@@ -63,7 +52,14 @@ func main() {
 		app.Stop()
 	}()
 
-	err = app.Run()
+	go app.Serve()
+
+	var err error
+	if cfg.App.ServerMode {
+		err = app.RunServer()
+	} else {
+		err = app.RunOnce()
+	}
 	logger := log.L()
 	if err != nil {
 		logger.Error("tidb lightning encountered error", zap.Error(err))
--- a/cmd/tidb-lightning/main.go
+++ b/cmd/tidb-lightning/main.go
@@ -23,19 +23,10 @@ import (
 	"github.com/pingcap/tidb-lightning/lightning"
 	"github.com/pingcap/tidb-lightning/lightning/config"
 	"github.com/pingcap/tidb-lightning/lightning/log"
-	plan "github.com/pingcap/tidb/planner/core"
 	"go.uber.org/zap"
 )
 
-func setGlobalVars() {
-	// hardcode it
-	plan.SetPreparedPlanCache(true)
-	plan.PreparedPlanCacheCapacity = 10
-}
-
 func main() {
-	setGlobalVars()
-
 	cfg := config.Must(config.LoadGlobalConfig(os.Args[1:], nil))
 	app := lightning.New(cfg)
 
--- a/lightning/config/configlist.go
+++ b/lightning/config/configlist.go
@@ -17,14 +17,14 @@ import (
 	"container/list"
 	"context"
 	"sync"
+	"time"
 )
 
 // ConfigList is a goroutine-safe FIFO list of *Config, which supports removal
 // from the middle. The list is not expected to be very long.
 type ConfigList struct {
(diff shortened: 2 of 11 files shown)


GregoryIan:
@july2993 @lonng PTAL

lightning/config/global.go
```diff
@@ -0,0 +1,183 @@
...
+	"github.com/pingcap/tidb-lightning/lightning/log"
+)
+
+type GlobalLightning struct {
+	log.Config
+	StatusAddr string `toml:"status-addr" json:"status-addr"`
+	ServerMode bool   `toml:"server-mode" json:"server-mode"`
+
+	// The legacy alias for setting "status-addr"
+	PProfPort int `toml:"pprof-port" json:"-"`
```
lonng: I think it's better to add some comments about the reason for ignoring the `PProfPort` field in the `JSON` configuration. Otherwise, we should not ignore it.

cmd/tidb-lightning/main.go
```diff
@@ -38,16 +36,7 @@ func setGlobalVars() {
 func main() {
 	setGlobalVars()
```
lonng: Should we remove this function? The `PreparedPlanCache` feature has some extra cost but seems we never rely on this feature.

lightning/lightning.go
```diff
@@ -32,67 +34,105 @@ import (
...
+				finalErr = err
+			}
+		}
+	}
+}
+
+var taskCfgRecorderKey struct{}
+
+func (l *Lightning) run(taskCfg *config.Config) error {
+	failpoint.Inject("SkipRunTask", func() {
```
lonng: We can change the failpoint clousure signature to remove the `failpoint.Return` call.
```go
	failpoint.Inject("SkipRunTask", func() error {
		if recorder, ok := l.ctx.Value(&taskCfgRecorderKey).(chan *config.Config); ok {
			recorder <- taskCfg
		}
		return nil
	})
```

lightning/lightning.go
```diff
@@ -32,67 +34,105 @@ import (
...
 	}
 
-	if cfg.App.ProfilePort > 0 {
-		go func() {
-			http.Handle("/metrics", promhttp.Handler())
-			err := http.ListenAndServe(fmt.Sprintf(":%d", cfg.App.ProfilePort), nil)
-			log.L().Info("stopped HTTP server", log.ShortError(err))
-		}()
-	}
+	runtime.GOMAXPROCS(runtime.NumCPU())
```
lonng: It's not necessary to call `runtime.GOMAXPROCS` manually.

lightning/lightning.go
```diff
@@ -105,6 +145,59 @@ func (l *Lightning) run() error {
...
+		http.Error(w, fmt.Sprintf("invalid task configuration: %v", err), http.StatusBadRequest)
+		return
+	}
+
+	select {
+	case cfgs <- cfg:
+		w.WriteHeader(http.StatusOK)
+		w.Write([]byte("task queued"))
+	default:
+		http.Error(w, "task queue overflowed, please try again later", http.StatusServiceUnavailable)
```
lonng: The execution flow will reach here if the channel was closed, do we need to change the error message to adapt it. (Although we don't close the channel in the current implementation)

lightning/lightning.go
```diff
@@ -32,67 +34,105 @@ import (
...
+	var finalErr error
+	l.taskCfgs = make(chan *config.Config, 64)
+
+	log.L().Info("Lightning server is running, post to /tasks to start an import task")
+
+	for {
+		select {
+		case <-l.ctx.Done():
+			return l.ctx.Err()
+		case task := <-l.taskCfgs:
```
july2993: seems nowhere will close l.taskCfgs now? and will not push nil(should not)

kennytm:
PTAL again @lonng @july2993 

Changed the queue data structure from a bare channel to a linked hash map to allow removing an item from the queue, and removed the artificial queue size limit. The output format of `/task` is changed to JSON.

GregoryIan:
@lonng PTAL again

GregoryIan:
it would be easier to integrate into TiDB, right?

lightning/config/configlist.go
```diff
@@ -0,0 +1,117 @@
...
+				res <- cfg
+				break
+			}
+			cl.cond.Wait()
+		}
+	}()
+
+	select {
+	case cfg := <-res:
+		return cfg, nil
```
lonng: We need to close the `channel` before returning.

kennytm: Closing a channel is not necessary https://stackoverflow.com/questions/8593645/is-it-ok-to-leave-a-channel-open

kennytm:
@GregoryIan as a plug-in embedding the whole app, probably just the same difficulty.

GregoryIan:
should we add task ID into some component (like checkpoint and metrics)?

kennytm:
Checkpoint no, metrics maybe, logs yes.

GregoryIan:
if we finish each job, I think we should clear checkpoint or initial new checkpoint with new task id

GregoryIan:
discuss with @kennytm, we would use `time` as task ID, and add task ID into checkpoint database/file, and log

GregoryIan:
LGTM

kennytm:
PTAL @GregoryIan

PR #201, opened 2019-07-02
Improve errors and logs on syntax error / conversion failure

kennytm:

[TOOL-1355]

1. When the parser encountered a syntax error, limit the content printed out to be 256 characters or less, to avoid flooding the log file.
2. Stop logging the entire row when a CAST AS failed, instead only log the affected value.
3. Replaced some `errors.Trace` by `errors.Annotatef` to indicate the error's provenance.

### Check List 

Tests 

 - Unit test

Side effects

Related changes


kennytm:
PTAL @july2993 @csuzhangxc @GregoryIan

kennytm:
/run-all-tests

FYI @csuzhangxc  I've added one commit 04e2d111431c5a4dee78154d2c42744b3bb843ed to reduce the "missing checkpoint file" log from Warn level to Info level so DBAs won't think it is a serious bad issue.

PTAL @leoppro

lightning/log/testlogger.go
```diff
@@ -0,0 +1,36 @@
...
+package log
+
+import (
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
+	"go.uber.org/zap/zaptest"
+)
+
+// MakeTestLogger creates a Logger instance which produces JSON logs.
+func MakeTestLogger() (Logger, *zaptest.Buffer) {
```
leoppro: Is the `MakeTestLogger` function only used in the test?
if so, maybe we don't export this function will be better.

kennytm: Yes it is only used for testing. But it is used in two different packages (`lightning/mydump/csv_parser_test.go` and `lightning/restore/restore_test.go`), so this function cannot be private.

leoppro:
LGTM

kennytm:
@leoppro @csuzhangxc Could either one of you approve the PR? Thanks.

PR #206, opened 2019-07-04
tests: add test case for simple partitioned tables

kennytm:

Verify pingcap/tidb#10796 works on Lightning.

Added an integrated test case involving partitioned tables.

### Check List 

Tests 

 - Integration test

Side effects

Related changes


kennytm:
PTAL @lance6716 @leoppro

lance6716:
LGTM

PR #208, opened 2019-07-10
checkpoints: remove node_id field and rename the schema on keep-after-success

kennytm:

Resolve the previous comment in https://github.com/pingcap/tidb-lightning/pull/198#issuecomment-502603038.

1. If `checkpoint.keep-after-success` is enabled, we will 
    * rename the checkpoint file to `{schema}.pb.{taskID}.bak` with `file` driver, or
    * rename the schema to `` `{schema}.{taskID}.bak` `` with `mysql` driver
2. The latter is actually incorrect when `node_id` is present. The proper way to "rename" when respecting `node_id` is to do a `CREATE TABLE` + `INSERT SELECT` + `DELETE`, which is a mess (TiDB doesn't support `CREATE TABLE SELECT` yet). Since we still haven't finalized the distributed import strategy, we take the opportunity here to allocate an entire schema to one node, and in case of multiple nodes they have to choose different schema names.
3. With this `node_id` change we changed many checkpoint management procedures. In particular the "--checkpoint-remove=all" operation now simply drops the entire schema.

### Check List 

Tests 

 - Unit test
 - Integration test

Side effects

    * The file name / database schema name after using `keep-after-success` is no longer constant.

Related changes


lightning/checkpoints/checkpoints.go
```diff
@@ -388,15 +389,13 @@ func NewMySQLCheckpointsDB(ctx context.Context, db *sql.DB, schemaName string) (
 
 	err = sql.Exec(ctx, "create table checkpoints table", fmt.Sprintf(`
 		CREATE TABLE IF NOT EXISTS %s.%s (
-			node_id int unsigned NOT NULL,
-			session bigint unsigned NOT NULL,
+			task_id bigint NOT NULL,
```
csuzhangxc: can we combine `task_id` and `table_name` as a UNIQUE KEY or is it needed?

kennytm: Not strictly needed, this is mainly for debugging currently.

lightning/checkpoints/checkpoints.go
```diff
@@ -457,17 +453,17 @@ func (cpdb *MySQLCheckpointsDB) Initialize(ctx context.Context, dbInfo map[strin
 
 	s := common.SQLWithRetry{DB: cpdb.db, Logger: log.L()}
 	err := s.Transact(ctx, "insert checkpoints", func(c context.Context, tx *sql.Tx) error {
-		// If `node_id` is not the same but the `table_name` duplicates,
+		// If `hash` is not the same but the `table_name` duplicates,
```
csuzhangxc: Is `hash` still always `0` now (in L477)?

lightning/checkpoints/checkpoints.go
```diff
@@ -937,57 +936,68 @@ func (*NullCheckpointsDB) DumpChunks(context.Context, io.Writer) error {
...
 
 func (cpdb *MySQLCheckpointsDB) IgnoreErrorCheckpoint(ctx context.Context, tableName string) error {
-	var (
-		colName string
-		arg     interface{}
-	)
+	var colName string
 	if tableName == "all" {
-		colName, arg = "node_id", nodeID
+		colName = "'all'"
```
csuzhangxc: oh, maybe you can add a comment here to make it clear, `WHERE 'all' = 'all'`

lightning/checkpoints/checkpoints.go
kennytm: For now still true

lightning/checkpoints/checkpoints.go
```diff
@@ -364,12 +365,12 @@ func (*NullCheckpointsDB) InsertEngineCheckpoints(_ context.Context, _ string, _
 func (*NullCheckpointsDB) Update(map[string]*TableCheckpointDiff) {}
 
 type MySQLCheckpointsDB struct {
-	db      *sql.DB
-	schema  string
-	session uint64
```
GregoryIan: any reason to remove session？ don't we use it forever?

lightning/checkpoints/checkpoints.go
```diff
@@ -388,15 +389,13 @@ func NewMySQLCheckpointsDB(ctx context.Context, db *sql.DB, schemaName string) (
...
-			node_id int unsigned NOT NULL,
-			session bigint unsigned NOT NULL,
+			task_id bigint NOT NULL,
 			table_name varchar(261) NOT NULL PRIMARY KEY,
 			hash binary(32) NOT NULL,
 			status tinyint unsigned DEFAULT 30,
 			alloc_base bigint NOT NULL DEFAULT 0,
 			create_time timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
-			update_time timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
-			INDEX(node_id, session)
```
GregoryIan: index `task_id`?

lightning/checkpoints/checkpoints.go
kennytm: For `node_id` it's explained in the PR description. 

For `session` it's just renamed to `task_id`.

GregoryIan: In fact I don't understand the description, why must need `create select`, `node id` can be just a label

lightning/checkpoints/checkpoints.go
```diff
@@ -937,57 +936,68 @@ func (*NullCheckpointsDB) DumpChunks(context.Context, io.Writer) error {
...
+		if _, e := tx.ExecContext(c, deleteTableQuery, tableName); e != nil {
 			return errors.Trace(e)
 		}
 		return nil
 	})
-	return errors.Trace(err)
+}
+
+func (cpdb *MySQLCheckpointsDB) MoveCheckpoints(ctx context.Context, taskID int64) error {
+	newSchema := fmt.Sprintf("`%s.%d.bak`", cpdb.schema[1:len(cpdb.schema)-1], taskID)
```
GregoryIan: is it better to use a function `unescapedSchemaName `?

GregoryIan:
Rest LGTM

lightning/checkpoints/checkpoints.go
kennytm: In the original design, multiple nodes are sharing the same table distinguished by node_id.

<table><tr><td>checkpoints.table_v5

| node_id | ... |
|---------|-----|
| 1       | A   |
| 1       | B   |
| 2       | C   |
| 2       | D   |

</td></tr></table>

To support keep-after-success + server mode, we'd want to extract the content like this:

<table><tr><td>checkpoints.table_v5

| node_id | ... |
|---------|-----|
| 2       | C   |
| 2       | D   |

</td><td>checkpoints_backup.table_v5

| node_id | ... |
|---------|-----|
| 1       | A   |
| 1       | B   |

</td></tr></table>

So that future checkpoints from node 1 won't interfere with the previous checkpoint, meanwhile the checkpoints of node 2 are not affected. 

The CREATE SELECT is needed to copy content to `checkpoints_backup.table_v5`.

lightning/checkpoints/checkpoints.go
kennytm: Added a comment explaining why unescaping isn't used (to keep the escaped `` ` ``s)

kennytm:
@csuzhangxc @GregoryIan PTAL again, thanks.

kennytm:
Ping @GregoryIan

GregoryIan:
I will take a look ASAP 😭

lightning/checkpoints/checkpoints.go
GregoryIan: If you have a distributed job mode in the future, do you still have to modify it? This modification is confusing, there may be many feasible solutions, but it's not key problem

kennytm:
(Need to fix the missing tikv-importer binary first, SRE problem)

PR #209, opened 2019-07-13
tests,web: exclude these directories from the Go module

kennytm:

Currently when trying to import tidb-lightning externally on macOS we'll get a "malformed file path" error due to a `` ` `` in the filename of a test case. Since a Go module shouldn't need the integrated tests anyway, we just ignore these non-source folders instead of trying to fix Go 🤷.

Recommended in https://github.com/golang/go/wiki/Modules#can-an-additional-gomod-exclude-unnecessary-content-do-modules-have-the-equivalent-of-a-gitignore-file, we include an empty `go.mod` in the `tests/` and `web/` folder so they're no longer part of the Go module.

### Check List 

Tests 

 - No code

Side effects

Related changes

### DIFF
--- a/tests/go.mod
+++ b/tests/go.mod
@@ -1,0 +1,1 @@
+// Exclude this directory from the Go module
--- a/web/go.mod
+++ b/web/go.mod
@@ -1,0 +1,1 @@
+// Exclude this directory from the Go module


kennytm:
PTAL @csuzhangxc @lance6716

lance6716:
LGTM

PR #217, opened 2019-07-23
config: attempt to solve TOOL-1405 and modify old test cases

lance6716:
Early exit to save time when user's toml config contains unused items, for example, a typo. And give warnings when some config items won't take effect.

A check is performed on `(*Config).LoadFromTOML`, using `toml.Decode` and `metaData.Undecoded` to detect unused toml keys. Considering we have two structure `Config` and `GlobalConfig`, we check for both unused keys and give according exit-action or warning.

### Check List 

Tests 

 - Unit test
 - Integration test

Side effects

Related changes


lance6716:
PTAL @kennytm @leoppro

lightning/config/config.go
```diff
@@ -16,6 +16,7 @@ package config
 import (
 	"encoding/json"
 	"fmt"
+	"go.uber.org/zap"
```
leoppro: code format , we usually import external dependencies separately

lightning/config/config.go
```diff
@@ -204,8 +205,58 @@ func (cfg *Config) LoadFromGlobal(global *GlobalConfig) error {
...
 // LoadFromTOML overwrites the current configuration by the TOML data
+// If data contains toml items not in Config and GlobalConfig, return an error
+// If data contains toml items not in Config, thus won't take effect, warn user
 func (cfg *Config) LoadFromTOML(data []byte) error {
-	return errors.Trace(toml.Unmarshal(data, cfg))
+	// warn legal toml items but won't effect
+	var warnItems []string
+	dataStr := string(data)
+	metaData, err := toml.Decode(dataStr, cfg)
+
```
leoppro: throw error here?

lightning/config/config.go
```diff
@@ -204,8 +205,58 @@ func (cfg *Config) LoadFromGlobal(global *GlobalConfig) error {
...
+	// Here's a warning that we haven't check err before use metaData,
+	// but metaData will always return a valid struct and we check err in next line
+	undecoded := metaData.Undecoded()
+	if len(undecoded) > 0 && err == nil {
```
leoppro: Short circuit logic better here

lightning/config/config.go
```diff
@@ -204,8 +205,58 @@ func (cfg *Config) LoadFromGlobal(global *GlobalConfig) error {
...
+		}
+	}
+
+	// Warn that some field of config file won't be overwrite, such as lightning.file
+	if len(warnItems) > 0 {
+		log.L().Warn("these fields of config file have seen twice but could not be overwritten, it's ok if user was not intended to change them",
+			zap.String("content", strings.Join(warnItems, ", ")))
+	}
+
+	return errors.Trace(err)
```
leoppro: add unit test

lightning/config/config.go
```diff
@@ -204,8 +205,58 @@ func (cfg *Config) LoadFromGlobal(global *GlobalConfig) error {
...
 func (cfg *Config) LoadFromTOML(data []byte) error {
-	return errors.Trace(toml.Unmarshal(data, cfg))
+	// warn legal toml items but won't effect
+	var warnItems []string
+	dataStr := string(data)
+	metaData, err := toml.Decode(dataStr, cfg)
+
+	// Here's a warning that we haven't check err before use metaData,
+	// but metaData will always return a valid struct and we check err in next line
+	undecoded := metaData.Undecoded()
```
leoppro: the variable name seems not good enough

lightning/config/config.go
lance6716: GoLand helps me 😭

lightning/config/config.go
lance6716: `err` is kept and returned at last lines of this function. I referenced [this piece](https://github.com/pingcap/tidb/blob/ddb6132e82f79a91518d75c91617f3d60d0de06b/config/config.go#L527-L538). But in fact now I prefer check `err` here.

lightning/config/config.go
lance6716: how about change `undecoded` to `unused`

leoppro:
LGTM

lightning/config/config.go
```diff
@@ -204,8 +205,68 @@ func (cfg *Config) LoadFromGlobal(global *GlobalConfig) error {
...
+	for _, keyStr := range unusedConfigKeyStrs {
+		if _, found := unusedGlobalKeyStrs[keyStr]; found {
+			bothUnused = append(bothUnused, keyStr)
+		} else {
+			warnItems = append(warnItems, keyStr)
+		}
+	}
+
+	if len(bothUnused) > 0 {
+		err = errors.New(fmt.Sprintf("config file contained unknown configuration options: %s",
```
kennytm:
```suggestion
		err = errors.Errorf("config file contained unknown configuration options: %s",
```

lightning/config/config.go
```diff
@@ -204,8 +205,68 @@ func (cfg *Config) LoadFromGlobal(global *GlobalConfig) error {
...
+			strings.Join(bothUnused, ", ")))
+		return errors.Trace(err)
```
kennytm: No need to `errors.Trace` any errors created using `errors.New` or `errors.Errorf`.

```suggestion
		return err
```

lightning/config/config.go
```diff
@@ -204,8 +205,68 @@ func (cfg *Config) LoadFromGlobal(global *GlobalConfig) error {
...
+	}
+
+	// Warn that some legal field of config file won't be overwritten, such as lightning.file
+	if len(warnItems) > 0 {
+		log.L().Warn("these fields of config file have seen more than once but could not be overwritten, it's ok if user was not intended to change them",
```
kennytm: I don't quite understand the meaning of the log message :)

lance6716: This warning was made for user who want to change `GlobalConfig`, such as `lightning.file`, by changing task-toml sent to server. Task-toml could only effect valid member of `Config`, not `GlobalConfig`. if program would silently ignore them, user may get confused.

But we use same toml file to construct `GlobalConfig` and `Config` in terminal, so in fact this warning will always appear in terminal-using situation. I'm not sure above case worth always writing a warning in log.

Maybe I could check `warnItems`' value, for if they are same with `GlobalConfig`, only warn when not-same.

lightning/config/config.go
```diff
@@ -204,8 +205,64 @@ func (cfg *Config) LoadFromGlobal(global *GlobalConfig) error {
...
+
+	if len(bothUnused) > 0 {
+		return errors.Errorf("config file contained unknown configuration options: %s",
+			strings.Join(bothUnused, ", "))
+	}
+
+	// Warn that some legal field of config file won't be overwritten, such as lightning.file
+	if len(warnItems) > 0 {
+		log.L().Warn("currently only per-task configuration can be applied, global configuration changes can only be made on startup",
+			zap.String("global config changes", strings.Join(warnItems, ", ")))
```
kennytm:
```suggestion
			zap.Strings("global config changes", warnItems))
```

PR #218, opened 2019-07-23
restore: fix gc life time not recovered after table restore

amyangfei:

feedback from DBA, sometimes after import data successfully with lightning, `tikv_gc_life_time` remains `100h0m0s` and not recovers to the original value (often `10m0s`).

This is caused by concurrent execution of `restore table`, and the `tikv_gc_life_time` update timeline maybe intersect like following:

```
thread1:  get gc=10m ---> set gc=100h -----> checksum ------> rollback gc to 10m
thread2:  -------------------------------> get gc=100h -----> checksum -----> rollback gc to 100h
```
We query `tikv_gc_life_time` from downstream database before execution of restore table once

### Check List 

Tests 

 - Unit test
 - Integration test

Related changes

### DIFF
--- a/lightning/restore/restore.go
+++ b/lightning/restore/restore.go
@@ -490,12 +490,17 @@ func (rc *RestoreController) restoreTables(ctx context.Context) error {
 	}
 	taskCh := make(chan task, rc.cfg.App.IndexConcurrency)
 	defer close(taskCh)
+	oriGCLifeTime, err := ObtainGCLifeTime(ctx, rc.tidbMgr.db)
+	if err != nil {
+		return errors.Trace(err)
+	}
+	ctx2 := context.WithValue(ctx, "tikv_gc_life_time", oriGCLifeTime)
 	for i := 0; i < rc.cfg.App.IndexConcurrency; i++ {
 		go func() {
 			for task := range taskCh {
 				tableLogTask := task.tr.logger.Begin(zap.InfoLevel, "restore table")
 				web.BroadcastTableCheckpoint(task.tr.tableName, task.cp)
-				err := task.tr.restoreTable(ctx, rc, task.cp)
+				err := task.tr.restoreTable(ctx2, rc, task.cp)
 				tableLogTask.End(zap.ErrorLevel, err)
 				web.BroadcastError(task.tr.tableName, err)
 				metric.RecordTableCount("completed", err)
@@ -541,7 +546,7 @@ func (rc *RestoreController) restoreTables(ctx context.Context) error {
 	wg.Wait()
 	stopPeriodicActions <- struct{}{}
 
-	err := restoreErr.Get()
+	err = restoreErr.Get()
 	logTask.End(zap.ErrorLevel, err)
 	return err
 }
@@ -1356,9 +1361,15 @@ func DoChecksum(ctx context.Context, db *sql.DB, table string) (*RemoteChecksum,
 func increaseGCLifeTime(ctx context.Context, db *sql.DB) (oriGCLifeTime string, err error) {
 	// checksum command usually takes a long time to execute,
 	// so here need to increase the gcLifeTime for single transaction.
-	oriGCLifeTime, err = ObtainGCLifeTime(ctx, db)
-	if err != nil {
-		return "", errors.Trace(err)
+	// try to get gcLifeTime from context first.
+	gcLifeTime, ok := ctx.Value("tikv_gc_life_time").(string)
+	if !ok {
+		oriGCLifeTime, err = ObtainGCLifeTime(ctx, db)
+		if err != nil {
+			return "", errors.Trace(err)
+		}
+	} else {
+		oriGCLifeTime = gcLifeTime
 	}
 
 	var increaseGCLifeTime bool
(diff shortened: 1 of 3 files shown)


tests/concurrent-restore/run.sh
```diff
@@ -0,0 +1,49 @@
...
+echo 'CREATE DATABASE restore_conc;' > "$DBPATH/restore_conc-schema-create.sql"
+for i in $(seq "$TABLE_COUNT"); do
+    echo "CREATE TABLE tbl$i(i TINYINT);" > "$DBPATH/restore_conc.tbl$i-schema.sql"
+    echo "INSERT INTO tbl$i VALUES (1);" > "$DBPATH/restore_conc.tbl$i.sql"
+done
+
+run_sql 'select VARIABLE_VALUE from mysql.tidb where VARIABLE_NAME = "tikv_gc_life_time"';
+ORIGINAL_TIKV_GC_LIFE_TIME=$(tail -n 1 "$TEST_DIR/sql_res.$TEST_NAME.txt" | awk '{print $(NF)}')
+
+# Count OpenEngine and CloseEngine events.
```
kennytm: Remove this line

lightning/restore/restore.go
```diff
@@ -490,12 +490,17 @@ func (rc *RestoreController) restoreTables(ctx context.Context) error {
 	}
 	taskCh := make(chan task, rc.cfg.App.IndexConcurrency)
 	defer close(taskCh)
+	oriGCLifeTime, err := ObtainGCLifeTime(ctx, rc.tidbMgr.db)
+	if err != nil {
+		return errors.Trace(err)
+	}
+	ctx2 := context.WithValue(ctx, "tikv_gc_life_time", oriGCLifeTime)
```
kennytm: https://godoc.org/context#WithValue:

> The provided key must be comparable and **should not be of type string** or any other built-in type to avoid collisions between packages using context.

Try using a global pointer like

```go
var gcLifeTimeKey struct{}

func (rc *RestoreController) restoreTables(ctx context.Context) error {
    ...
    ctx2 := context.WithValue(ctx, &gcLifeTimeKey, oriGCLifeTime)
    ...
}
```

lightning/restore/restore.go
```diff
@@ -490,12 +490,17 @@ func (rc *RestoreController) restoreTables(ctx context.Context) error {
 	}
 	taskCh := make(chan task, rc.cfg.App.IndexConcurrency)
 	defer close(taskCh)
+	oriGCLifeTime, err := ObtainGCLifeTime(ctx, rc.tidbMgr.db)
+	if err != nil {
+		return errors.Trace(err)
```
kennytm:
```suggestion
		return err
```

Let's not use unnecessary `errors.Trace` in new code 🙂

lightning/restore/restore.go
```diff
@@ -1356,9 +1361,15 @@ func DoChecksum(ctx context.Context, db *sql.DB, table string) (*RemoteChecksum,
 func increaseGCLifeTime(ctx context.Context, db *sql.DB) (oriGCLifeTime string, err error) {
 	// checksum command usually takes a long time to execute,
 	// so here need to increase the gcLifeTime for single transaction.
-	oriGCLifeTime, err = ObtainGCLifeTime(ctx, db)
-	if err != nil {
-		return "", errors.Trace(err)
+	// try to get gcLifeTime from context first.
+	gcLifeTime, ok := ctx.Value("tikv_gc_life_time").(string)
```
kennytm: (ditto)

tests/concurrent-restore/run.sh
```diff
@@ -0,0 +1,49 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -eu
+
+# Populate the mydumper source
+DBPATH="$TEST_DIR/restore.mydump"
+TABLE_COUNT=20
```
kennytm: Is the test reproducible using less number of tables and a static data source?

amyangfei: the more tables, the higher probability to reproduce this bug. I tested four tables seems enough to reproduce, I changed the table count to 8

lightning/restore/restore.go
```diff
@@ -490,12 +492,17 @@ func (rc *RestoreController) restoreTables(ctx context.Context) error {
 	}
 	taskCh := make(chan task, rc.cfg.App.IndexConcurrency)
 	defer close(taskCh)
+	oriGCLifeTime, err := ObtainGCLifeTime(ctx, rc.tidbMgr.db)
+	if err != nil {
+		return err
+	}
+	ctx2 := context.WithValue(ctx, gcLifeTimeKey, oriGCLifeTime)
```
kennytm: Err no you do need a *pointer* here. Every `struct{}` are the same but  `*struct{}` can be different.

```suggestion
	ctx2 := context.WithValue(ctx, &gcLifeTimeKey, oriGCLifeTime)
```

tests/concurrent-restore/run.sh
```diff
@@ -0,0 +1,48 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -eu
+
+# Populate the mydumper source
+DBPATH="$TEST_DIR/restore.mydump"
```
kennytm: Please rename the folder to something other than `restore.mydump` (nor `cppk.mydump` nor `cpch.mydump`).

PR #220, opened 2019-07-29
restore: update and restore GCLifeTime once when parallel

lance6716:
DoChecksum doesn't prepare for parallel case 

There may be multiple `DoChecksum` running, setting and resetting `GC Life Time` should not be done when there's unfinished `DoChecksum` tasks.

Lightning have `restoreTables` runs [simultaneously](https://github.com/pingcap/tidb-lightning/blob/master/lightning/lightning.go#L153), so we use this logic 

```go
func restoreTables() {
    // init_helper
    for some conurrency {
        go restoreTable()
    }
}

func restoreTable() {
    // call postProcess() -> ... -> DoChecksum()
}

func DoChecksum() {
    // using helper's pointers of lock, running jobs counter, original value
}
```
Using variable to count running checksum jobs. Before a remote chechsum call starting, lock, increase this counter, check if it just rised from zero then backup original value and call set `GC Life Time` logic, unlock.
After a remote chechsum finishing, lock, decrease this counter, check if it just drop to zero then call resetting `GC Life Time` logic, unlock.

The lock, counter, original value are pointed to same per location in one `restoreTables`.

### Check List 

Tests 

 - Unit test
 - Integration test

Side effects

Related changes


lance6716:
Seems that it's another solution of #218. Should I borrow the code and integration test and revert context-related code? @kennytm @amyangfei

kennytm:
🤔 Oh I thought you've found another test case which isn't fixed by #218.

An advantage of the `Context` based solution is that it doesn't rely on global variables.

lance6716:
> 🤔 Oh I thought you've found another test case which isn't fixed by #218.
> 
> An advantage of the `Context` based solution is that it doesn't rely on global variables.

I'll learn `Context` and fix that conflict tomorrow.

lance6716:
@leoppro found a bug in my atomic integer code.
```
C1:->atomic check---------------------------------------->set GC->
C2:--------------->atomic check--->remote checksum call---------->
```
We think there should be a mutex, so I'll add three variable: a mutex for set GC time, a atomic integer as job counter(or a mutex for reset GC time), original GC time.

I still prefer using global variable, since lightning can only deal with one TiDB, and three variables are about "lock for TiDB setting, TiDB running jobs, TiDB setting" accrodingly, global variable won't pollute a lot. 

Another way is using context like #218, assign three-variable-struct to context in `restoreTables` and get values in `DoChecksum`. 

I'd like to learn your suggestions.

kennytm:
@lance6716 use a struct, related variables should be placed close to each other.

lance6716:
PTAL @kennytm @amyangfei

lance6716:
PTAL @leoppro

lightning/restore/restore.go
```diff
@@ -1344,19 +1360,37 @@ func setSessionConcurrencyVars(ctx context.Context, db *sql.DB, dsn config.DBSto
 // DoChecksum do checksum for tables.
 // table should be in <db>.<table>, format.  e.g. foo.bar
 func DoChecksum(ctx context.Context, db *sql.DB, table string) (*RemoteChecksum, error) {
-	ori, err := increaseGCLifeTime(ctx, db)
-	if err != nil {
-		return nil, errors.Trace(err)
+	var err error
+	helper, ok := ctx.Value(&gcLifeTimeKey).(gcLifeTimeHelper)
+	if !ok {
+		return nil, errors.Errorf("No gcLifeTimeHelper found in context, check context initialization")
```
amyangfei:
```suggestion
		return nil, errors.New("No gcLifeTimeHelper found in context, check context initialization")
```

lightning/restore/restore.go
```diff
@@ -1344,19 +1360,37 @@ func setSessionConcurrencyVars(ctx context.Context, db *sql.DB, dsn config.DBSto
...
+	}
+
+	helper.increaseGCLock.Lock()
+	*(helper.runningJobs) += 1
+	if *(helper.runningJobs) == 1 {
+		err = increaseGCLifeTime(ctx, db)
+		if err != nil {
+			helper.increaseGCLock.Unlock()
+			return nil, errors.Trace(err)
```
amyangfei:
```suggestion
			return nil, err
```
error that returns from `increaseGCLifeTime` already has the stack trace

lightning/restore/restore.go
```diff
@@ -1412,13 +1441,13 @@ func increaseGCLifeTime(ctx context.Context, db *sql.DB) (oriGCLifeTime string,
 	if increaseGCLifeTime {
 		err = UpdateGCLifeTime(ctx, db, defaultGCLifeTime.String())
 		if err != nil {
-			return "", errors.Trace(err)
+			return errors.Trace(err)
```
amyangfei:
```suggestion
			return err
```

lightning/restore/restore.go
```diff
@@ -1344,19 +1360,37 @@ func setSessionConcurrencyVars(ctx context.Context, db *sql.DB, dsn config.DBSto
...
 	// set it back finally
 	defer func() {
-		err := UpdateGCLifeTime(ctx, db, ori)
-		if err != nil {
-			query := fmt.Sprintf("UPDATE mysql.tidb SET VARIABLE_VALUE = '%s' WHERE VARIABLE_NAME = 'tikv_gc_life_time'", ori)
-			log.L().Warn("revert GC lifetime failed, please reset the GC lifetime manually after Lightning completed",
-				zap.String("query", query),
-				log.ShortError(err),
-			)
+		if atomic.AddInt32(helper.runningJobs, -1) == 0 {
```
amyangfei: Can this scenario happen, the condition `atomic.AddInt32(helper.runningJobs, -1) == 0` in this line is true, but before L1383 executes, another RetstoreTable routine starts and increases `helper.runningJobs` by 1 ?

lightning/restore/restore.go
lance6716: Silly me 😂

lightning/restore/restore.go
lance6716: You're right.
In fact another RestoreTable routine has its own `helper.runningJobs`, 
[L507](https://github.com/pingcap/tidb-lightning/pull/220/files#diff-9697521ce51bfb4c9ef44a387e73c780R507). But two tightly running RestoreTable routines will have undetermined result.

Now I want to make `helper.runningJobs` point to one global integer which represents only one TiDB, and use same lock not atomic operations. Hence `gcLifeTimeLock` and `runningJobs` become file-global. What do you advise?

amyangfei: I'm not familiar with the `server mode` feature, but if the task is running one by one in the server mode, not in a concurrent way, a global file-level lock for `runningJobs` is fine.

lance6716:
/run-all-tests

I updated code and PR description. @amyangfei @leoppro

kennytm:
What if we change `DoChecksum` to a method of `(rc *RestoreController)`, and store the GCLifeTime lock stuff into the `rc`?

lance6716:
> What if we change `DoChecksum` to a method of `(rc *RestoreController)`, and store the GCLifeTime lock stuff into the `rc`?

I've noticed `RestoreController` contains some lock, but the function call is

`RestoreController.restoreTables` -> `TableRestore.restoreTable` -> `TableRestore.postProcess` -> `TableRestore.compareChecksum` -> `DoChecksum(context.Context, *sql.DB, string)`

`DoChecksum` seems can't reach `RestoreController` easily.

kennytm:
We could add an argument containing the locks to `compareChecksum` though

lance6716:
I don't understand. By current code `DoChecksum` routines should share a counter, a lock, a string, we could structure them as a global variable or pass them in Context. Passing argument seems similar to context-way, but they three will travel from `restoreTable` to `DoChecksum`, a long distance I think.

lightning/restore/restore.go
```diff
@@ -474,6 +472,66 @@ func (rc *RestoreController) runPeriodicActions(ctx context.Context, stop <-chan
 	}
 }
 
+type gcLifeTimeManager struct {
+	runningJobsLock *sync.Mutex
+	runningJobs     *int32
+	oriGCLifeTime   *string
```
kennytm: Instead of a struct of 3 pointers, use a pointer to a struct of 3 values instead.

```go
type gcLifeTimeManager struct {
    runningJobsLock sync.Mutex
    runningJobs     int   // no need to use int32 here, btw.
    oriGCLifeTime   string
}
```

lightning/restore/restore.go
```diff
@@ -474,6 +472,66 @@ func (rc *RestoreController) runPeriodicActions(ctx context.Context, stop <-chan
...
+}
+
+func newManager() gcLifeTimeManager {
```
kennytm: (and return a pointer here)

```suggestion
func newManager() *gcLifeTimeManager {
```

lightning/restore/restore.go
```diff
@@ -474,6 +472,58 @@ func (rc *RestoreController) runPeriodicActions(ctx context.Context, stop <-chan
...
+	m.runningJobs += 1
+	if m.runningJobs == 1 {
+		oriGCLifeTime, err := ObtainGCLifeTime(ctx, db)
+		if err != nil {
+			return err
+		}
+		m.oriGCLifeTime = oriGCLifeTime
+		err = increaseGCLifeTime(ctx, db)
+		if err != nil {
+			return err
```
kennytm: If any of these branches returned an error (thus `increaseGCLifeTime` didn't take effect), the `runningJobs` would still have value "1"...

lightning/restore/restore.go
```diff
@@ -1344,21 +1392,18 @@ func setSessionConcurrencyVars(ctx context.Context, db *sql.DB, dsn config.DBSto
...
-		err := UpdateGCLifeTime(ctx, db, ori)
-		if err != nil {
-			query := fmt.Sprintf("UPDATE mysql.tidb SET VARIABLE_VALUE = '%s' WHERE VARIABLE_NAME = 'tikv_gc_life_time'", ori)
-			log.L().Warn("revert GC lifetime failed, please reset the GC lifetime manually after Lightning completed",
-				zap.String("query", query),
-				log.ShortError(err),
-			)
-		}
-	}()
+	defer manager.removeOneJob(ctx, db)
```
kennytm: ... but this `removeOneJob` won't be called due to early return on line 1402. 

I suggest you move line 490 (`m.runningJob += 1`) _after_ `increaseGCLifeTime` takes effect. 

Also, if possible, add a test case to verify the behavior of failing `increaseGCLifeTime`?

lance6716: Ah so precisely. I'm going to add this comment to `addOneJob` to remind me of this mistake:
```
// Pre- and post-condition:
// if m.runningJobs == 0, GC life time has not been increased.
// if m.runningJobs > 0, GC life time has been increased.
// m.runningJobs won't be negative(overflow) since index concurrency is relatively small
```

And this to `removeOneJob`:
```
// Pre- and post-condition:
// if m.runningJobs == 0, GC life time has been tried to recovered. If this try fails, a warning will be printed.
// if m.runningJobs > 0, GC life time has not been recovered.
// m.runningJobs won't minus to negative since removeOneJob follows a successful addOneJob.
```

lightning/restore/restore.go
```diff
@@ -474,6 +472,66 @@ func (rc *RestoreController) runPeriodicActions(ctx context.Context, stop <-chan
 	}
 }
 
+type gcLifeTimeManager struct {
+	runningJobsLock sync.Mutex
+	runningJobs     int
+	oriGCLifeTime   string
+}
+
+func newManager() *gcLifeTimeManager {
```
kennytm: Oh also please rename this method to something more specific, since it is seen by the whole package 😅. 

```suggestion
func newGCLifeTimeManager() *gcLifeTimeManager {
```

lightning/restore/restore_test.go
```diff
@@ -148,6 +157,88 @@ func (s *restoreSuite) TestDoChecksum(c *C) {
...
+			_, err = DoChecksum(ctx, db, "`test`.`t`")
+			c.Assert(err, ErrorMatches, "update GC lifetime failed: update gc error: context canceled")
+			wg.Done()
+		}()
+	}
+	wg.Wait()
+
+	// validate no more redundant GC recover
+	err = mock.ExpectationsWereMet()
+	c.Assert(err, ErrorMatches, "there is a remaining expectation.*\n.*\n.*\n.*\n.*\n.*\n.*")
```
kennytm: 😕 How about just don't add those remaining expectations?

lance6716: I tested that with this mock database, `db.Exec("some unexpected SQL")` will not fail. I guess if I didn't specify there's a remaining expectations, it won't be detected.

My test is removing `ExpectExec` of UPDATE and adding `db.Exec` of this update at end, and this test passes.

kennytm: @lance6716 That's odd, because if there's remaining expectations [it _will_ fail](https://github.com/DATA-DOG/go-sqlmock/blob/c05e0c0b9787ee10ae6f1ee6f5a5a70323bd6fcc/sqlmock.go#L174). 

Perhaps add an `ExpectClose()` and close the database instead?

lance6716: @kennytm Unexpected exec fails on the exec itself, with no effect on following `mock.ExpectationsWereMet()`'s success. If `DoChecksum` restores GC Life Time as an unexpected exec and doesn't reveal this error(perhaps this restore error is overwritten by update error), unit test can't detect that.

Of course by peeking into `DoChecksum` source code, we know unexpected exec occures in `defer` thus won't be overwritten, but I think unit test should not rely on source code check.

lightning/restore/restore_test.go
```diff
@@ -232,17 +236,16 @@ func (s *restoreSuite) TestIncreaseGCLifeTimeFail(c *C) {
...
 
-	// validate no more redundant GC recover
-	err = mock.ExpectationsWereMet()
-	c.Assert(err, ErrorMatches, "there is a remaining expectation.*\n.*\n.*\n.*\n.*\n.*\n.*")
+	_, err = db.Exec("\\QUPDATE mysql.tidb SET VARIABLE_VALUE = ? WHERE VARIABLE_NAME = 'tikv_gc_life_time'\\E", "10m")
+	c.Assert(err, IsNil)
 
-	mock.ExpectClose()
+	c.Assert(db.Close(), IsNil)
+	err = mock.ExpectationsWereMet()
```
kennytm: Assert it is not nil ?

lance6716: `db.Close()` returns `err`, so it's short for error check.

BTW I noticed L243 didn't have a check...

kennytm: @lance6716 sorry typo 😅, I do mean line 243 should be asserted to be nil.

lance6716:
/run-all-tests

@leoppro PTAL

kennytm:
Ping @leoppro

lightning/restore/restore.go
```diff
@@ -474,6 +472,66 @@ func (rc *RestoreController) runPeriodicActions(ctx context.Context, stop <-chan
...
+// if m.runningJobs > 0, GC life time has been increased.
+// m.runningJobs won't be negative(overflow) since index concurrency is relatively small
+func (m *gcLifeTimeManager) addOneJob(ctx context.Context, db *sql.DB) error {
+	m.runningJobsLock.Lock()
+	defer m.runningJobsLock.Unlock()
+
+	if m.runningJobs == 0 {
+		oriGCLifeTime, err := ObtainGCLifeTime(ctx, db)
+		if err != nil {
+			return err
```
leoppro: why we don't trace the error

lightning/restore/restore.go
```diff
@@ -1344,21 +1400,18 @@ func setSessionConcurrencyVars(ctx context.Context, db *sql.DB, dsn config.DBSto
...
-	ori, err := increaseGCLifeTime(ctx, db)
-	if err != nil {
-		return nil, errors.Trace(err)
+	var err error
+	manager, ok := ctx.Value(&gcLifeTimeKey).(*gcLifeTimeManager)
+	if !ok {
+		return nil, errors.New("No gcLifeTimeManager found in context, check context initialization")
+	}
+
+	if err = manager.addOneJob(ctx, db); err != nil {
```
leoppro: how about we rename `addOneJob` to `start` and rename `removeOneJob` to `done`.

lightning/restore/restore.go
lance6716: deep digging into this error, it shows this is an `Annotatef()`,  which has stack trace already

lightning/restore/restore.go
lance6716: ok start/done or start/finish? I prefer later

leoppro: I prefer `done`, but `finish` is fine

lance6716: sorry I don't support renaming now. `start` seems the job action is IN this function, but in fact `addOneJob` is intend to do prepare jobs, increase counter, e.g. Job action (checksum) is after the `addOneJob` call

PR #225, opened 2019-08-15
config: always skip the system databases

kennytm:

Avoid importing the built-in databases (`mysql`, `information_schema`, `performance_schema` and `sys`).

Add these 4 databases into the black list in `Adjust()`.

### Check List 

Tests 

 - Integration test

Side effects

    * User has no way to explicitly restore the tables defined inside `mysql`.

Related changes


lance6716:
LGTM

and I think user can restore those tables because

> If a database’s name matches both the do-dbs and ignore-dbs arrays, the database is included.

kennytm:
PTAL @csuzhangxc

ericsyh:
@kennytm Pls pick to the 3.0 branch

kennytm:
It's already in the 3.0 branch.

PR #227, opened 2019-08-18
backend: dynamically calculate the maximum auto-inc ID

### LINKED ISSUE #222: why lightning change table auto_increment?为什么lightning会修改自增值的大小？
log:
[INFO] [restore.go:1253] ["import and cleanup engine completed"] [engineTag=`test`.`favfav`:-1] [engineUUID=9c824252-d150-5902-a2c0-02d2f26ebf6c] [takeTime=8.515710692s] []
[INFO] [tidb.go:232] ["alter table auto_increment start"] [table=`test`.`favfav`] [auto_increment=1567203009]

question:
why lightning change the  auto_increment value before checksum? And the new value will be so big that would lead to long gap between max(primary id) and next new rows.
lightning 在导入数据结束，checksum前会修改自增值，如果表数据很大，新的自增值和实际的当前最大ID间的距离会非常大，有字段溢出的风险。

kennytm:

Fix #222 (= TOOL-1472). 

Remove the piece of code that pre-calculates the final AUTO_INCREMENT ID. Instead, let the encoder determines the actual value dynamically. This gives a tighter upper bound of the final AUTO_INCREMENT value and avoids overflowing AUTO_INCREMENT columns of shorter integer size (this doesn't work if PRIMARY KEY is not a handle column though, since `_tidb_rowid` will share the same AUTO_INCREMENT sequence)

### Check List 

Tests 

 - Integration test

Side effects

Related changes

### DIFF
--- a/lightning/backend/sql2kv.go
+++ b/lightning/backend/sql2kv.go
@@ -164,12 +164,13 @@ func (kvcodec *tableKVEncoder) Encode(
 
 	for i, col := range cols {
 		j := columnPermutation[i]
+		isAutoIncCol := mysql.HasAutoIncrementFlag(col.Flag)
 		if j >= 0 && j < len(row) {
 			value, err = table.CastValue(kvcodec.se, row[j], col.ToInfo())
 			if err == nil {
 				value, err = col.HandleBadNull(value, kvcodec.se.vars.StmtCtx)
 			}
-		} else if mysql.HasAutoIncrementFlag(col.Flag) {
+		} else if isAutoIncCol {
 			// we still need a conversion, e.g. to catch overflow with a TINYINT column.
 			value, err = table.CastValue(kvcodec.se, types.NewIntDatum(rowID), col.ToInfo())
 		} else {
@@ -179,6 +180,9 @@ func (kvcodec *tableKVEncoder) Encode(
 			return nil, logKVConvertFailed(logger, row, j, col.ToInfo(), err)
 		}
 		record = append(record, value)
+		if isAutoIncCol {
+			kvcodec.tbl.RebaseAutoID(kvcodec.se, value.GetInt64(), false)
+		}
 	}
 
 	if !kvcodec.tbl.Meta().PKIsHandle {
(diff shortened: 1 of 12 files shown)


lightning/backend/sql2kv.go
```diff
@@ -179,6 +180,9 @@ func (kvcodec *tableKVEncoder) Encode(
 			return nil, logKVConvertFailed(logger, row, j, col.ToInfo(), err)
 		}
 		record = append(record, value)
+		if isAutoIncCol {
+			kvcodec.tbl.RebaseAutoID(kvcodec.se, value.GetInt64(), false)
```
lance6716: This `RebaseAutoID` is called at every row in my understanding, why could this 
rebase-per-row gives tighter upper bound than rebase-per-chunk, I think they will eventually reach almost same auto-id and now it's more expensive.

kennytm: The change allowing a tighter bound is deleting

https://github.com/pingcap/tidb-lightning/blob/4eb74ec2a454e10c09113d7da89394cc22c417eb/lightning/restore/restore.go#L649-L653

since `RowIDMax` is normally much larger than the actual row number. 

Yes this could become more expensive, but we should perform a measurement to ensure. I bet the CAS loop is still much cheaper than the actual KV encoding.

lance6716: Now i see rebased value is cast from `rowId`, which is smaller or equal than row's content. And I agree CAS is lightweight. I don't know if we could move this rebase-per-row outside, such as on [chunk's EOF](https://github.com/pingcap/tidb-lightning/blob/master/lightning/restore/restore.go#L1660), so there's fewer repeat.

kennytm:
PTAL @lance6716 @WangXiangUSTC

lightning/backend/sql2kv.go
lance6716: CAS is lightweight so maybe we shouldn't bother move it outside ⬆️

lance6716:
LGTM

PR #230, opened 2019-08-23
cmd: do not exit(1) if failed to sync log

kennytm:

Even if the log file failed to sync (usually caused by log file = /dev/stdout), it shouldn't cause the program to exit with failure.

Remove the condition that sync log fail causing exit(1).

### Check List 

Tests 

    * Run the program

Side effects

Related changes

 - Need to cherry-pick to the release branch (2.1)

### DIFF
--- a/cmd/tidb-lightning/main.go
+++ b/cmd/tidb-lightning/main.go
@@ -66,7 +66,7 @@ func main() {
 		fmt.Fprintln(os.Stderr, "sync log failed", syncErr)
 	}
 
-	if err != nil || syncErr != nil {
+	if err != nil {
 		os.Exit(1)
 	}
 }


kennytm:
/run-all-tests

PTAL @lance6716 @WangXiangUSTC

lance6716:
LGTM

kennytm:
Thanks. But I'd argue that even if the error is _not_ an `*os.PathError`, we should still not cause the program to `exit(1)`, because the error code is used to determine whether the data is successfully imported into the TiDB cluster.  The log being incomplete is irrelevant if the import succeed.

gregwebs:
I agree. But I still want to know about any issues where the system is not behaving properly. So I would at least want to see the error logged (if it is not a PathError about stdout/stderr).

lance6716:
exclude minimal error 😄

kennytm:
@gregwebs The sync error is already written to stderr in lines 64–67.

gregwebs:
okay, that's good then. It is still going to produce spurious stderr though, so its still best to ignore it if we know specifically what the error is.

gregwebs:
It's confusing to still see the sync error in the log

Issue #231, opened 2019-08-23
Progress webpage fails to load in non-server mode


tennix:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do? If possible, provide a recipe for reproducing the error.

Deploying tidb-lightning using k8s via https://github.com/pingcap/tidb-operator/pull/817
```
git clone https://github.com/tennix/tidb-operator -b tidb-lightning
cd tidb-operator
helm install charts/tidb-cluster -n <cluster-release-name> --namespace <namepsace> --set importer.create=true
helm install charts/tidb-lightning -n <lightning-release-name> --namespace <namespace> --set dataSource.local.nodeName=<node-name>,dataSource.local.hostPath=<host-path>,targetTidbCluster.name=<cluster-release-name>
```
Then view the lightning service

2. What did you expect to see?

The web page shows the task progress correctly.

3. What did you see instead?

The page shows and disappears immediately.

4. Versions of the cluster

The version of the cluster is v3.0.1. The corresponding images are pulled from DockerHub.

PR #232, opened 2019-08-23
[WIP] lightning: print message on non-server mode

lance6716:
To clarify #231. Print friendly message when access Web API on non-server mode.

Add logic to check if it's server mode.

### Check List 

Tests 

   - By manually open 127.0.0.1:`pprof-port`, thinking of an integration test. 

Side effects

Related changes


kennytm:
Actually the web page should still be accessible in non-server-mode, just that the "queue task" function won't work. The web page should be fixed instead.

PR #235, opened 2019-08-29
*: use fixed timestamp to ensure import stability wrt CURRENT_TIMESTAMP

### LINKED ISSUE #171: Override the CURRENT_TIMESTAMP and store in checkpoint
After #170, Lightning can recognize the `DEFAULT CURRENT_TIMESTAMP` option, which is always [initialized to `time.Now()`](https://github.com/pingcap/tidb/blob/154fc6013cd63ce4a66748d30ab6f5d26e8722e3/expression/helper.go#L137). This means restoring from checkpoint is going to be non-deterministic when the timestamp column is indexed. 

```sql
CREATE TABLE t (
    x DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP,
    KEY(x)
);
INSERT INTO t VALUES (), (), (), ();
```

We need to override the session system variable `timestamp`, and also store it into the checkpoint to ensure the value before and after are equivalent.

kennytm:

Fix #171  

Before this PR, the CURRENT_TIMESTAMP value is evaluated to the actual wall-clock time. This means writing a row containing CURRENT_TIMESTAMP would not be idempotent. This would potentially break an index based on these columns.

Store the timestamp where a file is getting imported into the checkpoint. Use the `@@TIMESTAMP` system variable to pin the value of CURRENT_TIMESTAMP.

### Check List 

Tests 

 - Unit test
 - Integration test

Side effects

Related changes


lightning/backend/sql2kv_test.go
```diff
@@ -119,6 +119,36 @@ func (s *kvSuite) TestEncode(c *C) {
...
+
+	logger := log.Logger{Logger: zap.NewNop()}
+
+	encoder := NewTableKVEncoder(tbl, mysql.ModeStrictAllTables, 1234567893)
+	pairs, err := encoder.Encode(logger, nil, 70, []int{-1, 1})
+	c.Assert(err, IsNil)
+	c.Assert(pairs, DeepEquals, kvPairs([]kvenc.KvPair{
+		{
+			Key: []uint8{0x74, 0x80, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x1, 0x5f, 0x72, 0x80, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x46},
+			Val: []uint8{0x8, 0x2, 0x9, 0x80, 0x80, 0x80, 0xf0, 0xfd, 0x8e, 0xf7, 0xc0, 0x19},
```
lance6716: How did this value derived 🤔

lightning/backend/session.go
```diff
@@ -66,6 +68,7 @@ func newSession(sqlMode mysql.SQLMode) *session {
 	vars.StmtCtx.AllowInvalidDate = sqlMode.HasAllowInvalidDatesMode()
 	vars.StmtCtx.IgnoreZeroInDate = !sqlMode.HasStrictMode() || sqlMode.HasAllowInvalidDatesMode()
 	vars.StmtCtx.TimeZone = vars.Location()
+	vars.SetSystemVar("timestamp", strconv.FormatInt(timestamp, 10))
```
lance6716: What's the usage of this SystemVar, I didn't find usage in lightnong or doc of TiDB

kennytm: It's from MySQL. https://dev.mysql.com/doc/refman/8.0/en/server-system-variables.html#sysvar_timestamp

lightning/backend/sql2kv_test.go
kennytm: Through TiDB's `EncodeValue` function (though in reality I just get the actual output and paste it back, as the purpose of this test is to ensure we get the same fixed value everytime).

lance6716: Ah, fixed value is a good test point.
If this reproduce step is short we could paste those steps here and let code generates this value

lance6716:
LGTM

tests/checkpoint_timestamp/run.sh
```diff
@@ -0,0 +1,46 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -eu
+
+run_sql 'DROP DATABASE IF EXISTS cpts'
+rm -f "$TEST_DIR"/cpts.pb*
+
+export GO_FAILPOINTS="github.com/pingcap/tidb-lightning/lightning/SetTaskID=return(1234567890);github.com/pingcap/tidb-lightning/lightning/restore/FailIfImportedChunk=return(28)"
```
GregoryIan: `tests/checkpoint_timestamp/data/cpts.cpts.1.sql` is always in one chunk, I think we should split into two chunks to verify the correctness

GregoryIan:
Rest LGTM

kennytm:
/run-all-tests

PTAL @GregoryIan

GregoryIan:
CI was broken @kennytm

kennytm:
PTAL @GregoryIan

PR #236, opened 2019-08-30
backend, restore: Strip the first 11 bytes from key in WriteEngine, specify it in OpenEngine

lance6716:
Solve https://github.com/pingcap/kvproto/pull/451

For partition table, pipeline keeps same with original.

For non-partition table
1. at `OpenEngine`, lightning specify `recordPrefix` or `indexPrefix`; 
2. at `engine.WriteRows`, use a parameter to indicate prefix removal; 
3. at `rows.SplitIntoChunks`, strip each 11-byte-long prefix, to reduce `WriteEngine` data transmission.

### Check List 

Tests 

 - Unit test
 - integration test with [importer#23](https://github.com/tikv/importer/pull/23)

Side effects

 - Increased code complexity

Related changes


lance6716:
Seems that we could strip key prefix earlier, to reduce 11 bytes checksum computing. Instead, `(*KVChecksum).UpdateOne` will have a pre-computed prefix CRC value as initial value.

lance6716:
/run-all-tests

PTAL @kennytm @WangXiangUSTC

kennytm:
As concluded in tikv/importer#23, the key rewrite rule will no longer be specified in OpenEngine. The key-stripping algorithm should still work the same though, and thus I'm keeping this PR open until the new key-rewrite mechanism is in place.

GregoryIan:
can we close this pr firstly, open it while we need it? @kennytm

kennytm:
(Since we're going to merge the sorting part of Importer into Lightning, this transformation no longer has any advantages.)

PR #237, opened 2019-09-06
checkpoint: fix empty map become nil after unmarshall

lance6716:
fix empty map become nil after marshall and unmarshall

manually init these map. Seems it's a bug of marshall function provided by gogoproto.

### Check List 

Tests 

 - Unit test, add an unit test. If we didn't init these map after unmarshall this test won't pass(hope I didn't make it wrong)

Side effects

Related changes

### DIFF
--- a/lightning/checkpoints/checkpoints.go
+++ b/lightning/checkpoints/checkpoints.go
@@ -741,7 +741,25 @@ func NewFileCheckpointsDB(path string) *FileCheckpointsDB {
 	// ignore all errors -- file maybe not created yet (and it is fine).
 	content, err := ioutil.ReadFile(path)
 	if err == nil {
-		cpdb.checkpoints.Unmarshal(content)
+		err2 := cpdb.checkpoints.Unmarshal(content)
+		if err2 != nil {
+			log.L().Error("checkpoint file is broken", zap.String("path", path), zap.Error(err2))
+		}
+		// FIXME: patch for empty map may need initialize manually, because currently
+		// FIXME: a map of zero size -> marshall -> unmarshall -> become nil, see checkpoint_test.go
+		if cpdb.checkpoints.Checkpoints == nil {
+			cpdb.checkpoints.Checkpoints = map[string]*TableCheckpointModel{}
+		}
+		for _, table := range cpdb.checkpoints.Checkpoints {
+			if (*table).Engines == nil {
+				(*table).Engines = map[int32]*EngineCheckpointModel{}
+			}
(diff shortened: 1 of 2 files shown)


lance6716:
/run-all-tests

PTAL @kennytm @GregoryIan 

BTW patch branch for Tencent PCG is [here](https://github.com/pingcap/tidb-lightning/compare/release-2.1...pcg-debug) and cherry-pick to 2.1 will be based on this pcg-debug

lightning/checkpoints/checkpoints.go
```diff
@@ -741,7 +741,25 @@ func NewFileCheckpointsDB(path string) *FileCheckpointsDB {
 	// ignore all errors -- file maybe not created yet (and it is fine).
 	content, err := ioutil.ReadFile(path)
 	if err == nil {
-		cpdb.checkpoints.Unmarshal(content)
+		err2 := cpdb.checkpoints.Unmarshal(content)
+		if err2 != nil {
+			log.L().Error("checkpoint file is broken", zap.String("path", path), zap.Error(err2))
+		}
+		// FIXME: patch for empty map may need initialize manually, because currently
+		// FIXME: a map of zero size -> marshall -> unmarshall -> become nil, see checkpoint_test.go
```
GregoryIan: why using FIXME for these comments?

GregoryIan:
Rest LGTM

lightning/checkpoints/checkpoints_test.go
```diff
@@ -288,3 +288,17 @@ func (s *checkpointSuite) TestApplyDiff(c *C) {
 		},
 	})
 }
+
+func (s *checkpointSuite) TestCheckpointMarshallUnmarshall(c *C) {
+	path := "/tmp/test-chkp"
```
kennytm: Please use `c.MkDir()` and place the file inside that directory.

lightning/checkpoints/checkpoints.go
lance6716: I think it's an unexpected behaviour of `gogo/protobuf`'s marshall and unmarshall. There's logic to recover a empty map(but didn't execute) in https://github.com/pingcap/tidb-lightning/blob/master/lightning/checkpoints/file_checkpoints.pb.go#L877-L878

If `gogo/protobuf` fixed this bug, we won't bother to initiate. So I added a FIXME to remind to report bug and check if fixed from `gogo/protobuf`.

GregoryIan:
LGTM

Issue #242, opened 2019-09-25
Misc improvements


lance6716:
## Feature Request

**Is your feature request related to a problem? Please describe:**

**Describe the feature you'd like:**

- [ ] autoretry on io.EOF? because io.EOF is [an ambigious error](https://github.com/grpc/grpc-go/blob/230def769105c46b2e597578b23002a6ac159dde/stream.go#L96-L99)

- [ ] check if on error shouldn't [save checkpoint](https://github.com/pingcap/tidb-lightning/blob/aaac2b48c3a2954b2c8f6cc1de3d56c1bc41de8e/lightning/restore/restore.go#L911-L914) (but do importer's engine files remain?)

**Describe alternatives you've considered:**

**Teachability, Documentation, Adoption, Optimization:**

GregoryIan:
Good Catch

PR #243, opened 2019-09-26
backend/tidb: use REPLACE INTO or INSERT IGNORE INTO to provide idempotent insertion

lance6716:
1 modifications from #242 : 
- use `REPLACE INTO` to provide idempotent insertion while PK or UNIQUE exsits

replace according SQL to `REPLACE INTO` or `INSERT IGNORE INTO` statement in backend/tidb's WriteRows. 
add a config field to decide behaviour

```
[tikv-importer]
backend = "tidb"
on-duplicate = "replace" # or "ignore", "error". which represents REPLACE INTO, INSERT IGNORE INTO, INSERT INTO, respectively.
```

### Check List 

Tests 

 - Unit test
 - Integration test

Side effects

 - Possible performance regression: not sure `REPLACE INTO` is slower

Related changes

### DIFF
--- a/lightning/backend/tidb.go
+++ b/lightning/backend/tidb.go
@@ -52,6 +52,12 @@ type tidbBackend struct {
 // The backend does not take ownership of `db`. Caller should close `db`
 // manually after the backend expired.
 func NewTiDBBackend(db *sql.DB, onDuplicate string) Backend {
+	switch onDuplicate {
+	case config.ReplaceOnDup, config.IgnoreOnDup, config.ErrorOnDup:
+	default:
+		log.L().Warn("unsupported action on duplicate, overwrite with `replace`")
+		onDuplicate = config.ReplaceOnDup
+	}
 	return MakeBackend(&tidbBackend{db: db, onDuplicate: onDuplicate})
 }
 
--- a/lightning/backend/tidb.go
+++ b/lightning/backend/tidb.go
@@ -43,7 +43,7 @@ type tidbEncoder struct {
 
 type tidbBackend struct {
 	db *sql.DB
-	OnDuplicate string
+	onDuplicate string
 }
 
 // NewTiDBBackend creates a new TiDB backend using the given database.
--- a/lightning/backend/tidb.go
+++ b/lightning/backend/tidb.go
@@ -17,12 +17,11 @@ import (
 	"context"
 	"database/sql"
 	"encoding/hex"
-	"github.com/pingcap/failpoint"
-	"github.com/pingcap/tidb-lightning/lightning/config"
 	"strconv"
 	"strings"
 	"time"
 
+	"github.com/pingcap/failpoint"
 	"github.com/pingcap/parser/mysql"
 	"github.com/pingcap/tidb/table"
 	"github.com/pingcap/tidb/types"
@@ -30,6 +29,7 @@ import (
 	"go.uber.org/zap"
 
 	"github.com/pingcap/tidb-lightning/lightning/common"
+	"github.com/pingcap/tidb-lightning/lightning/config"
 	"github.com/pingcap/tidb-lightning/lightning/log"
 	"github.com/pingcap/tidb-lightning/lightning/verification"
 )
@@ -304,7 +304,7 @@ func (be *tidbBackend) WriteRows(ctx context.Context, _ uuid.UUID, tableName str
 	// Retry will be done externally, so we're not going to retry here.
 	_, err := be.db.ExecContext(ctx, insertStmt.String())
 	failpoint.Inject("FailIfImportedSomeRows", func() {
-		panic("forcing failure due to FailIfImportedSomeRows, before save checkpoint")
+		panic("forcing failure due to FailIfImportedSomeRows, before saving checkpoint")
 	})
 	return err
 }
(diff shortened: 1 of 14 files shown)


lance6716:
/run-all-tests

PTAL @kennytm @3pointer

kennytm:
Could we make it configurable to choose whether to use `INSERT INTO` / `REPLACE INTO` or `INSERT IGNORE INTO`?

lance6716:
Did a simple test using `mysql < text.sql`, `insert ignore into` is slightly faster(50% duplicated), so how about switch to `insert ignore into`?

ignore is reasonable since dumped files won't change.

kennytm:
This would mean that, during incremental import, if the new data does conflict with old data the default behavior is the new data being discarded, which is not very intuitive.

I think the default should still be REPLACE but we could configure to INSERT or IGNORE if needed

lance6716:
Oh I forget we should keep `INSERT INTO` as an option

GregoryIan:
LGTM

GregoryIan:
@kennytm PTAL

lance6716:
> @kennytm PTAL

Kenny says we might provide 3 types of behavior: replace, insert ignore, insert. I'm agree that we might provide all 3 types, and if a further benchmark shows there's little performance difference, we might reduce to 2 or 1 type at that time.

GregoryIan:
would lightning compute checksum many times for the rows which are executed more than twice? Rest LGTM

kennytm:
@GregoryIan Lightning currently won't do checksum with TiDB backend (since the ADMIN CHECKSUM is based on KV pairs).

kennytm:
@lance6716 I'll start review after you've changed the config from a bool to a string option.

lance6716:
/run-all-tests

PTAL @kennytm

lightning/backend/tidb.go
```diff
@@ -263,7 +265,15 @@ func (be *tidbBackend) WriteRows(ctx context.Context, _ uuid.UUID, tableName str
 	}
 
 	var insertStmt strings.Builder
-	insertStmt.WriteString("INSERT INTO ")
+	switch be.OnDuplicate {
```
GregoryIan: what's the default behavior of  `OnDuplicate`?

GregoryIan: should we give it a default value if users don't configure `OnDuplicate`

lance6716: default behavior is replacing on duplicate, which specifies in `Adjust()`.

https://github.com/pingcap/tidb-lightning/pull/243/files#diff-5f8541d99c1c66dcc6e24f4418ce5c24R205

`Adjust()` will be called in `RunOnce` and `handlePostTask()` (which then goes to RunServer), cover all user's case.

Codes call `NewTiDBBackend` should specify this param manually 🤔

tests/tidb_duplicate_data/run.sh
```diff
@@ -0,0 +1,40 @@
...
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -eu
+
```
GregoryIan: In fact we need such a case more:
* lightning start to loading
* panic on some statements (inject failpoint)
* restart lightning to finish loading
* check everything is ok

lightning/backend/tidb.go
```diff
@@ -42,14 +43,15 @@ type tidbEncoder struct {
 
 type tidbBackend struct {
 	db *sql.DB
+	OnDuplicate string
```
kennytm: Does any code outside this module needs to access this field? If not make the field private.

```suggestion
	onDuplicate string
```

(also, `go fmt` this file).

lightning/backend/tidb.go
```diff
@@ -42,14 +43,15 @@ type tidbEncoder struct {
...
 }
 
 // NewTiDBBackend creates a new TiDB backend using the given database.
 //
 // The backend does not take ownership of `db`. Caller should close `db`
 // manually after the backend expired.
-func NewTiDBBackend(db *sql.DB) Backend {
-	return MakeBackend(&tidbBackend{db: db})
+func NewTiDBBackend(db *sql.DB, OnDuplicate string) Backend {
```
kennytm:
```suggestion
func NewTiDBBackend(db *sql.DB, onDuplicate string) Backend {
```

Variable names should never be in uppercase.

lightning/config/config.go
```diff
@@ -47,6 +47,13 @@ const (
 	CheckpointDriverMySQL = "mysql"
 	// CheckpointDriverFile is a constant for choosing the "File" checkpoint driver in the configuration.
 	CheckpointDriverFile = "file"
+
+	// RepalceOnDup indicates using REPLACE INTO to insert data
+	RepalceOnDup = "replace"
```
kennytm:
```suggestion
	ReplaceOnDup = "replace"
```

lightning/backend/tidb.go
lance6716: my apologies, I simply delete prefix of `replaceOnDuplicate` and forget Go's semantic 😭

lance6716:
/run-all-tests

PTAL @kennytm @GregoryIan

lightning/backend/tidb.go
kennytm: IMO the extra defense is redundant. Not a big issue anyway.

PR #244, opened 2019-09-26
config: increase default concurrency

lance6716:
To reduce config of TiDB backend and user's effort, indexConcurrency and tableConcurrency may set to a large value automatically.

Computing the sum of tables to be imported, set indexConcurrency and tableConcurrency to this sum in TiDB backend mode

### Check List 

Tests 

 - Unit test

Side effects

Related changes

### DIFF
--- a/lightning/config/config.go
+++ b/lightning/config/config.go
@@ -47,10 +47,6 @@ const (
 	CheckpointDriverMySQL = "mysql"
 	// CheckpointDriverFile is a constant for choosing the "File" checkpoint driver in the configuration.
 	CheckpointDriverFile = "file"
-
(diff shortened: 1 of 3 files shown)


lance6716:
/run-all-tests

PTAL @kennytm @3pointer

lance6716:
`Adjust()` is a better place if we don't need to calculate number-of-tables.

How about setting them to region-concurrency? I'm not awared of performance impact if they're too big (calculating chunk distribution is simple), but seems indeed strange.

kennytm:
> How about setting them to region-concurrency?

Sounds good to me.

PR #248, opened 2019-11-19
config: adding `[tidb] max-allowed-packet` config

kennytm:
this allows sending rows > 4 MiB when using TiDB backend

Expose the [`maxAllowedPacket` DSN parameter](https://github.com/go-sql-driver/mysql/#maxallowedpacket) to `tidb-lightning.toml` (perhaps we should just accept arbitrary DSNs 🤷‍♂)

* Accept the `[tidb] max-allowed-packet = 67_108_864` config.
* When constructing the DSN to TiDB, include this value into the parameters.

### Check List 

Tests 

 - Unit test

Side effects

Related changes

 - Need to update the documentation (document the new config)
 - Need to update the `tidb-ansible` repository (add the new config to the templates)
 - Need to be included in the release note (?)

### DIFF
--- a/lightning/common/util.go
+++ b/lightning/common/util.go
@@ -46,12 +46,12 @@ const (
 	defaultMaxRetry = 3
 )
 
-func ToDSN(host string, port int, user string, psw string, sqlMode string) string {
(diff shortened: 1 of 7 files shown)


kennytm:
PTAL @3pointer @GregoryIan

WangXiangUSTC:
LGTM

PR #249, opened 2019-12-11
Fix +incompatible suffix not allowed error reported by go mod

suzaku:

Fix +incompatible suffix not allowed error reported by go mod

1. `go get -ugithub.com/pingcap/tidb@694e086e7`  (which is v3.0.4)
1. `go mod tidy`

### Check List 

Tests 

 - Manual test
    Use `make lightning` to check if we can still build.
 - No code

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -16,7 +16,7 @@ require (
 	github.com/pingcap/kvproto v0.0.0-20191011042334-8ee4fd8fb4ca
 	github.com/pingcap/log v0.0.0-20190307075452-bd41d9273596
 	github.com/pingcap/parser v0.0.0-20190910041007-2a177b291004
-	github.com/pingcap/tidb v3.0.4+incompatible
+	github.com/pingcap/tidb v1.1.0-beta.0.20190929123532-694e086e7914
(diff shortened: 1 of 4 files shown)


kennytm:
No need to get rid of `testify`?

suzaku:
> No need to get rid of `testify`?

I haven't figured out how to handle `testify` yet, it was added back to `go.mod` after `go mod tidy`.

kennytm:
@suzaku perhaps try to delete `go.sum` and run `go mod tidy` again

suzaku:
> @suzaku perhaps try to delete `go.sum` and run `go mod tidy` again

It's still there as before, with version 1.2.2.

kennytm:
I mean delete both `go.sum` and the `testify` line in `go.mod`, then run `go mod tidy`?

suzaku:
> I mean delete both `go.sum` and the `testify` line in `go.mod`, then run `go mod tidy`?

Yes, I've tried that too and the result was the same.

kennytm:
Let's take care of that later.

PR #253, opened 2019-12-16
Add password as command line argument

jlerche:
Previously it is not possible to supply the user password by command line, unlike other restore tools like `loader`

Adds `-tidb-password` command line option

### Check List 

Tests 

 - Unit test

Side effects

 - Increased code complexity

Related changes

### DIFF
--- a/lightning/config/config.go
+++ b/lightning/config/config.go
@@ -222,6 +222,7 @@ func (cfg *Config) LoadFromGlobal(global *GlobalConfig) error {
 	cfg.TiDB.Host = global.TiDB.Host
 	cfg.TiDB.Port = global.TiDB.Port
 	cfg.TiDB.User = global.TiDB.User
+	cfg.TiDB.Psw  = global.TiDB.Psw
 	cfg.TiDB.StatusPort = global.TiDB.StatusPort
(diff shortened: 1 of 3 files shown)


lightning/config/config.go
```diff
@@ -222,6 +222,7 @@ func (cfg *Config) LoadFromGlobal(global *GlobalConfig) error {
 	cfg.TiDB.Host = global.TiDB.Host
 	cfg.TiDB.Port = global.TiDB.Port
 	cfg.TiDB.User = global.TiDB.User
+	cfg.TiDB.Psw  = global.TiDB.Psw
```
gregwebs: I am more familiar with "Pwd" as an abbreviation, but do we need to abbreviate this at all?

jlerche: I used `Psw` to mimic the field name in the config

tennix: What did you mean by "mimic the field name in the config"? There is no `Psw` field in the configuration file.

kennytm: The field can be named "Psw" for consistency with old code (or all renamed to "Password"), but the CLI argument should be named `-tidb-password` not `-tidb-psw`.

jlerche:
https://github.com/pingcap/docs/pull/1711 PR for updating docs

PR #274, opened 2020-02-28
backend: define a dummy BufStore when creating a new session

kennytm:

In pingcap/tidb#12466 the `LightningMode` session variable was removed. One side effect is that a new 4 KB `BufStore` was initialized for every call to `AddRecord()`, which incurs 9% of the `Encode()` run time.

This PR constructs a dummy `BufStore` which forwards all writes to the transaction directly, similar to how `LightningMode` was implemented. 

<details><summary>Flamegraphs of the benchmark function against the TPC-C CUSTOMER table</summary>

Before: encode-only-profile-before-fs8

After: 
Annotation 2020-02-29 001955-fs8

</details>

### Check List 

Tests 

 - Unit test

Side effects

Related changes

### DIFF
--- a/lightning/backend/session.go
+++ b/lightning/backend/session.go
@@ -24,6 +24,20 @@ import (
 	"github.com/pingcap/tidb/sessionctx/variable"
 )
 
+// invalidIterator is a trimmed down Iterator type which is invalid.
+type invalidIterator struct {
+	kv.Iterator
+}
+
+// Valid implements the kv.Iterator interface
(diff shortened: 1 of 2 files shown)


kennytm:
PTAL @3pointer @bobotu cc @siddontang @XuHuaiyu

kennytm:
For record, the time needed to write 33 GB of the entire TPC-C "CUSTOMER" CSV table is reduced from 3m20.952730187s to 2m48.379138906s (84% of original).

PR #281, opened 2020-03-10
optimize the performance of lightning

july2993:
optimize the performance of lightning

the main performance improve is by setting `SetGCPercent as `500` default` compare to `Use pool for mutation` and `Reuse slice of record`. `disable UpdateDeltaForTable if TxnCtx is nil` should reduce 14% time in `Encode` according to the work by kennytm before in the comment.
- set SetGCPercent as `500` default

Lightning allocates too many transient objects and heap size is small,
so garbage collections happen too frequently and lots of time is spent in GC component.

In a test of loading the table `order_line.csv` of 14k TPCC.
The time need for `encode kv data and write` step reduce from 52m4s to 37m30s when change
GOGC from 100 to 500, the total time needed to restore the table reduce near 15m too.
The cost of this is the memory of lightning at runtime grows from about 200M to 700M, but it's acceptable.

So we set the GC percentage as 500 default to reduce the GC frequency instead of 100.

  -  Use pool for mutation

see commit https://github.com/pingcap/tidb-lightning/pull/281/commits/cce3ea65de77926ea127e6ae2c90ab79fa472b9a

-  Reuse slice of record
 https://github.com/pingcap/tidb-lightning/pull/281/commits/6f21d82f1e6ff10a66058201d3ff7963f904ebce
see commit 

- update dependency version of tidb
commit: https://github.com/pingcap/tidb-lightning/pull/281/commits/0181a3f7c8c736e24f00d2a86d96a6278a86789a
For pingcap/tidb@495f8b7
disable UpdateDeltaForTable if TxnCtx is nil

### Check List 

Tests 

 - Unit test
 - Integration test

Side effects

Related changes


lightning/mydump/csv_parser.go
```diff
@@ -50,6 +50,8 @@ type CSVParser struct {
 	// fieldIndexes is an index of fields inside recordBuffer.
 	// The i'th field ends at offset fieldIndexes[i] in recordBuffer.
 	fieldIndexes []int
+
+	lastRecord []string
```
kennytm: This field is used nowhere.

lightning/restore/restore.go
```diff
@@ -948,7 +948,7 @@ func (t *TableRestore) restoreEngine(
 	totalSQLSize := int64(0)
 	for _, chunk := range cp.Chunks {
 		totalKVSize += chunk.Checksum.SumSize()
-		totalSQLSize += chunk.Chunk.EndOffset
+		totalSQLSize += chunk.Chunk.EndOffset // want EndOffset - Offset ?
```
kennytm: Yes

lightning/restore/restore.go
```diff
@@ -691,7 +691,7 @@ func (rc *RestoreController) restoreTables(ctx context.Context) error {
 	}
 
 	wg.Wait()
-	stopPeriodicActions <- struct{}{}
+	close(stopPeriodicActions)
```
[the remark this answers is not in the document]
july2993: NO

init in `restoreTables` and close in `restoreTables` once
`runPeriodicActions` only consume it.
```
➜  tidb-lightning git:(refine_code) ✗ ag stopPeriodicActions
lightning/restore/restore.go
584:    stopPeriodicActions := make(chan struct{})
585:    go rc.runPeriodicActions(ctx, stopPeriodicActions)
694:    close(stopPeriodicActions)
➜  tidb-lightning git:(refine_code) ✗ vi lightning/restore/restore.go
```

lightning/restore/restore.go
july2993: updated

lightning/mydump/csv_parser.go
july2993: maybe removed when merging master and resolve conflict before...
revert back: https://github.com/pingcap/tidb-lightning/pull/281/commits/aa83ba885ef47e44bbc2635f42f74b26f87fda4a

lightning/backend/session.go
```diff
@@ -60,9 +60,12 @@ func (t *transaction) Iter(k kv.Key, upperBound kv.Key) (kv.Iterator, error) {
 
 // Set implements the kv.Mutator interface
 func (t *transaction) Set(k kv.Key, v []byte) error {
+	val := make([]byte, len(v))
+	copy(val, v)
+
 	t.kvPairs = append(t.kvPairs, common.KvPair{
 		Key: k.Clone(),
-		Val: append([]byte{}, v...),
+		Val: val,
```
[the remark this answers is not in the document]
july2993: remain append: https://github.com/pingcap/tidb-lightning/pull/281/commits/430f66a54c796071bb2491a440e47bd34554c2af
benchmark tells it will not faster..

july2993:
@kennytm PTAL

PR #296, opened 2020-04-20
restore: fix tablename case issue

AndrewDi:
tablename and schema could be upper case,we should not convert it to lower case

remove toLower func

### Check List 

Tests 

 - Unit test

Side effects

 - Increased code complexity

Related changes


kennytm:
/run-all-tests tidb=release-4.0 tikv=release-4.0 pd=release-4.0

kennytm:
I don't think this fix is relevant to the issue...

glorv:
since we have supported `mydumper.case-sensitive`, shall we close this now?

kennytm:
yes let's close.

thanks for the contribution @AndrewDi 🙂.

PR #304, opened 2020-05-08
restore: fix a typo about the recommended action on checkpoint error.

kennytm:

The suggested checkpoint error resolution command should be `--checkpoint-error-destroy=all`. Instead it was suggested to be `--checkpoint-errors-destroy=all`. There should be no `s`.

### Check List 

Tests 

 - Integration test

Side effects

Related changes


kennytm:
PTAL @YuJuncen

YuJuncen:
LGTM

Issue #306, opened 2020-05-09
Support automatically skipping version check when pd-url is not given


lichunzhu:
## Feature Request

**Is your feature request related to a problem? Please describe:**

**Describe the feature you'd like:**

Set `-check-requirements=false` when pd-url is not given to avoid failure.

PR #314, opened 2020-05-11
skip version check when pd-addr is not given

lichunzhu:
https://github.com/pingcap/tidb-lightning/issues/306
Sometimes user doesn't give pd-addr but the default `CheckRequirements` is true and version check will fail.

Skip version check but warn users when pd-addr is not give but `CheckRequirements` is true.

### Check List 

Tests 

 - Unit test
 - Integration test


lichunzhu:
https://github.com/pingcap/tidb-lightning/pull/312 will fix this issue. PdAddr will be adjusted so this judgement is useless.

Issue #309, opened 2020-05-09
warn users when the single loaded file is too big(has not been splitted by mydumper/dumpling)


lichunzhu:
## Feature Request

**Is your feature request related to a problem? Please describe:**

Single large loaded file will affect the effiency of `tidb-lightning`.

**Describe the feature you'd like:**

warning user when the single loaded file is too big(has not been splitted by mydumper/dumpling)

PR #315, opened 2020-05-11
Warn for single large file, change switch mode log level to info

lichunzhu:
https://github.com/pingcap/tidb-lightning/issues/309
Warn users when single table region size exceeds 1G.

Add warning log.

### Check List 

Tests 

 - Unit test
 - Integration test

### DIFF
--- a/lightning/mydump/region.go
+++ b/lightning/mydump/region.go
@@ -190,8 +190,10 @@ func MakeTableRegions(
 		}
 		filesRegions = append(filesRegions, tableRegion)
 		if tableRegion.Size() > tableRegionSizeWarningThreshold {
-			log.L().Warn("single sql file's size exceeds 1G, please modify dump configs to split file" +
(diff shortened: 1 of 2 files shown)


lightning/mydump/region.go
```diff
@@ -182,7 +187,12 @@ func MakeTableRegions(
 				PrevRowIDMax: prevRowIDMax,
 				RowIDMax:     rowIDMax,
 			},
-		})
+		}
+		filesRegions = append(filesRegions, tableRegion)
+		if tableRegion.Size() > tableRegionSizeWarningThreshold {
+			log.L().Warn("single sql file's size exceeds 1G, please modify dump configs to split file" +
+				" in several smaller files to improve lightning's efficiency", zap.String("file", dataFile))
```
kennytm, line 194:
```suggestion
			log.L().Warn(
				"file is too big to be processed efficiently; we suggest splitting it at 256 MB each",
				zap.String("file", dataFile),
				zap.Int64("size", dataFileSize))
```

PR #316, opened 2020-05-11
config: remove strict mode from default SQL mode

kennytm:

Relax the default SQL mode to remove the following strict mode components:

* STRICT_TRANS_TABLES
* NO_ZERO_IN_DATE
* NO_ZERO_DATE
* ERROR_FOR_DIVISION_BY_ZERO
* NO_AUTO_CREATE_USER (said to be deprecated in MySQL docs)

The remaining enabled modes are:

* ONLY_FULL_GROUP_BY
* NO_ENGINE_SUBSTITUTION

### Check List 

Tests 

 - Integration test

Side effects

    * the default SQL mode is changed.

Related changes

    * Lightning no longer assumes strict SQL mode when importing data for better compatibility. If you need to enforce strict mode, please change the `[tidb] sql-mode` configuration.


kennytm:
Should we *enable* [ALLOW_INVALID_DATES](https://dev.mysql.com/doc/refman/5.7/en/sql-mode.html#sqlmode_allow_invalid_dates) by default?

PR #324, opened 2020-05-27
support alter random && update tidb dependency to latest

### LINKED ISSUE #298: Support rebase auto_random allocator
## Bug Report

Currently when TiDB open auto_random feature in 3.1 or later version. lightning won't rebase auto_random allocator after restore data. then it may cause `Duplicate Entry` error in later transactions.

auto_random allocator rebase when restore finished, just like auto_increment.

3. What did you see instead?

4. Versions of the cluster

    - TiDB-Lightning version (run `tidb-lightning -V`):

        ```
        all version
        ```
        
    - TiKV-Importer version (run `tikv-importer -V`)

        ```
        all version
        ```
        
    - TiKV version (run `tikv-server -V`):

        ```
        all version
        ```

    - TiDB cluster version (execute `SELECT tidb_version();` in a MySQL client):

        ```
        3.1 or later
        ```

3pointer:
Fix #298 

1. Rebase increment part of auto random id in every kv record.
2. Reset auto random by alter auto_random_base query.

### Check List 

Tests 

Related changes


lightning/backend/session.go
```diff
@@ -112,8 +112,6 @@ func newSession(options *SessionOptions) *session {
 		txn:  transaction{},
 		vars: vars,
 	}
-	s.vars.GetWriteStmtBufs().BufStore = &kv.BufferStore{MemBuffer: &s.txn}
```
kennytm, line 115: This line is used to avoid some allocation and improve performance. Does the updated TiDB not support this?

3pointer: Updated TiDB removed this value, maybe there is other way to support it. I'll check it.

kennytm: Perhaps just check if the flamegraph leading to #274 still happens.

3pointer:
I found that after tidb use internel membuf in [AddRecord](https://github.com/pingcap/tidb/blame/master/table/tables/tables.go#L515), it's hard for us to inject our [logic](https://github.com/pingcap/tidb-lightning/blob/98bc8495f3fd2e74da20ffd623786f3839469d20/lightning/backend/session.go#L62) into function, because it will [discard](https://github.com/pingcap/tidb/blame/master/table/tables/tables.go#L515) membuffer immediately after AddRecord.

kennytm:
OK let's just ignore that issue at the moment.

.gitignore
```diff
@@ -12,3 +12,4 @@ tools/bin/
 # for the web interface
 web/node_modules/
 web/dist/
+*DS_Store
```
kennytm, line 15: Remove this line (the one in #326 is better)

lightning/restore/restore.go
```diff
@@ -747,8 +747,13 @@ func (t *TableRestore) restoreTable(
 		web.BroadcastTableCheckpoint(t.tableName, cp)
 
 		// rebase the allocator so it exceeds the number of rows.
-		cp.AllocBase = mathutil.MaxInt64(cp.AllocBase, t.tableInfo.Core.AutoIncID)
-		t.alloc[0].Rebase(t.tableInfo.ID, cp.AllocBase, false)
+		if t.tableInfo.Core.PKIsHandle && t.tableInfo.Core.ContainsAutoRandomBits() {
+			cp.AllocBase = mathutil.MaxInt64(cp.AllocBase, t.tableInfo.Core.AutoRandID)
+			t.alloc[2].Rebase(t.tableInfo.ID, cp.AllocBase, false)
```
kennytm, line 752: the "2" looks pretty magical. could you test the allocator's type instead?

lightning/backend/session.go
```diff
@@ -38,11 +45,184 @@ func (*invalidIterator) Valid() bool {
...
-	kvPairs []common.KvPair
+	kvPairs     []common.KvPair
+	bufferStore *memDbBuffer
+}
+
+func (t *transaction) NewStagingBuffer() kv.MemBuffer {
+	if t.bufferStore == nil {
+		t.bufferStore = NewMemDbBuffer()
+	}
+	return t.bufferStore
```
kennytm, line 225: why can't we use `t` as the MemBuffer?

3pointer: because of [this](https://github.com/pingcap/tidb-lightning/pull/324/files#diff-ea68acd8c8c86c13a06db5cb0b8bd058R180-R190) , current tidb will discard staging kv before `AddRecord` return, so our `memDbBuffer` is different from `kv.MemBuffer`,  it has `RealDiscard` method to do real discard.

kennytm: we could implement `(*transaction).Discard()` to do nothing?

lightning/restore/restore.go
3pointer: should I retain the old logic? only alloc[0] is enough, because all alloc shared the same base in one table

kennytm: no need to retain the old logic.

kennytm:
PTAL @YuJuncen

3pointer:
run-cherry-picker

PR #326, opened 2020-06-01
backend: add local kv storage backend to get ride of importer

glorv:
Use local key-value storage as a new backend to get ride of the dependency of `tikv-importer`, thus make lightning easier to use. 
In our benchmark, the performance of import speed with `local` mode is as good as `importer` mode, so this change won't bring any performance loss. Thus it will be a much better choice in compare with `tidb` backend.

The logic of `local` mode is as follows:
1. Write the data read for csv/mydumper to local kv-value storage `[pebble](https://github.com/cockroachdb/pebble)`.
2. Batch write sorted kv pairs and generate sst file at each tikv instance. https://github.com/tikv/tikv/pull/7459
3. ingest the sst file to tikv cluster

In the local mode, because sorted kv data are managed by the lightning process, so there are some change for `checkpoint` in local mode:
- In the restore phase with checkpoint enabled, if we restore an engine after restart, we will reset all the chunks within each engine to the initial state. This is because when lightning exited before engine closed, there maybe some data written to kv store but not flushed, thus these data may lost. Another approach is to do a flush after each chunk processed, but this is so much slow. 
- Before we update an engine checkpoint to `CheckpointStatusClosed`, we will do a flush to the related index-engine to make sure related index kvs are saved.
- Skip the `CheckpointStatusAllWritten` stage for local mode, because at the point , the data/index kv are not flushed, so if lightning exits at this point, we can't promise local kv store contains all the key-values for this engine, thus we have to restore the engine from start.
- Add a new meta file for each engine store with the engine db files. This meta containes some data used in the import phase. The meta file is generated when engine is closed.

Changes for `tidb-lightning-ctl`:
- the `import-engine` command is not supported in `local` mode. Because now import phase is done  by lightning, so this command is not meanful anymore.

NOTE:
- We recommend to separate the `sorted-kv-dir` from the data disk if possible, because the data-disk si read heavy, and the local storage dir is both read and write heavy, use another disk for this temp-store will make at least 10% performance gain.

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Side effects

Related changes


go.mod
```diff
@@ -5,23 +5,29 @@ go 1.13
...
-	github.com/pingcap/parser v0.0.0-20200317021010-cd90cc2a7d87
-	github.com/pingcap/tidb v0.0.0-20200317142013-5268094afe05
-	github.com/pingcap/tidb-tools v4.0.0-rc.1.0.20200514040632-f76b3e428e19+incompatible
+	github.com/pingcap/parser v0.0.0-20200331080149-8dce7a46a199
+	github.com/pingcap/pd/v4 v4.0.0-beta.1.0.20200305072537-61d9f9cc35d3
+	github.com/pingcap/tidb v1.1.0-beta.0.20200402143739-827edde58535
+	github.com/pingcap/tidb-tools v4.0.0-beta.1.0.20200306084441-875bd09aa3d5+incompatible
 	github.com/prometheus/client_golang v1.5.1
 	github.com/prometheus/client_model v0.2.0
 	github.com/satori/go.uuid v1.2.0
```
kennytm, line 28: we shouldn't need two UUID libraries.

go.mod
```diff
@@ -5,23 +5,29 @@ go 1.13
...
 	github.com/shurcooL/httpgzip v0.0.0-20190720172056-320755c1c1b0
+	github.com/syndtr/goleveldb v1.0.1-0.20190625010220-02440ea7a285
```
kennytm, line 30: is goleveldb used somewhere?

go.mod
```diff
@@ -5,23 +5,29 @@ go 1.13
 require (
 	github.com/BurntSushi/toml v0.3.1
 	github.com/DATA-DOG/go-sqlmock v1.4.1
+	github.com/cockroachdb/pebble v0.0.0-20200508213501-5a133cfa147f
 	github.com/coreos/go-semver v0.3.0
+	github.com/dgraph-io/badger/v2 v2.0.3
```
kennytm, line 10: is badger used somewhere?

lightning/backend/local.go
```diff
@@ -0,0 +1,1159 @@
...
+		v.(*LocalFile).Close()
+		return true
+	})
+
+	if !local.checkpointEnabled || common.IsEmptyDir(local.localStoreDir) {
+		err := os.RemoveAll(local.localStoreDir)
+		if err != nil {
+			log.L().Error("remove local db file failed", zap.Error(err))
+		}
+	}
```
kennytm, line 224:
```suggestion
	if !local.checkpointEnabled {
		err := os.Remove(local.localStoreDir)
		if err != nil {
			log.L().Warn("remove local db file failed", zap.String("dir", local.localStoreDir), log.ShortError(err))
		}
	}
```

* Don't use Error level for non-fatal errors.
* Is it normal that we close the local engine with left over files? I guess a warning in case of non-empty directory is fine.

lightning/backend/local.go
```diff
@@ -0,0 +1,1159 @@
...
+func (local *local) Flush(engineId uuid.UUID) error {
+	if engine, ok := local.engines.Load(engineId); ok {
+		engineFile := engine.(*LocalFile)
+		if err := engineFile.db.Flush(); err != nil {
+			return err
+		}
+		return local.saveEngineMeta(engineFile)
+	} else {
+		return errors.Errorf("engine '%s' not found", engineId.String())
+	}
```
kennytm, line 236:
```suggestion
	}
	return errors.Errorf("engine '%s' not found", engineId)
```

lightning/backend/local.go
```diff
@@ -0,0 +1,1159 @@
...
+		MaxConcurrentCompactions: 16,
+		MinCompactionRate:        1024 << 20,
+		L0CompactionThreshold:    2 << 31, // set to max try to disable compaction
+		L0StopWritesThreshold:    2 << 31, // set to max try to disable compaction
+		MaxOpenFiles:             10000,
+		DisableWAL:               true,
+	}
+	if readOnly {
+		opt.ReadOnly = true
+	}
```
kennytm, line 264:
```suggestion
		ReadOnly:                 readOnly,
	}
```

lightning/backend/local.go
```diff
@@ -0,0 +1,1159 @@
...
+	mu   sync.Mutex
+	clis map[uint64]*grpc.ClientConn
+}
+
+type local struct {
+	engines  sync.Map
+	grpcClis grpcClis
+	splitCli split.SplitClient
+	tls  *common.TLS
+	pdAddr 	string
```
kennytm, line 115: this file needs `gofmt`.

lightning/backend/local.go
```diff
@@ -0,0 +1,1159 @@
...
+	if err != nil {
+		return nil, errors.Trace(err)
+	}
+	opt := grpc.WithInsecure()
+	if local.tls.TransToTlsConfig() != nil {
+		opt = grpc.WithTransportCredentials(credentials.NewTLS(local.tls.TransToTlsConfig()))
+	}
+	ctx, cancel := context.WithTimeout(ctx, dialTimeout)
+	keepAlive := 10
+	keepAliveTimeout := 3
```
kennytm, line 189:
```suggestion
	keepAlive := 10 * time.Second
	keepAliveTimeout := 3 * time.Second
```

we better, move these to the const section.

lightning/backend/local.go
```diff
@@ -0,0 +1,1159 @@
...
+		MinCompactionRate:        1024 << 20,
+		L0CompactionThreshold:    2 << 31, // set to max try to disable compaction
+		L0StopWritesThreshold:    2 << 31, // set to max try to disable compaction
+		MaxOpenFiles:             10000,
+		DisableWAL:               true,
+	}
+	if readOnly {
+		opt.ReadOnly = true
+	}
+	dbPath := path.Join(local.localStoreDir, engineUUID.String())
```
kennytm, line 265:
```suggestion
	dbPath := filepath.Join(local.localStoreDir, engineUUID.String())
```

lightning/backend/local.go
```diff
@@ -0,0 +1,1159 @@
...
+	var meta localFileMeta
+	if err != nil {
+		return meta, err
+	}
+	fileBytes, err := ioutil.ReadAll(f)
+	if err != nil {
+		return meta, err
+	}
+
+	err = json.Unmarshal(fileBytes, &meta)
```
kennytm, line 295: Either use 

* `ioutil.ReadFile` + `json.Unmarhsal`, or
* `os.Open` + `json.NewDecoder(f).Decode(&meta)`

lightning/backend/local.go
```diff
@@ -0,0 +1,1159 @@
...
+		size += int64(len(iter.Key()) + len(iter.Value()))
+		pair := &sst.Pair{
+			Key:   bytesBuf.addBytes(iter.Key()),
+			Value: bytesBuf.addBytes(iter.Value()),
+			//Key: append([]byte{}, iter.Key()...),
+			//Value:append([]byte{}, iter.Value()...),
+		}
+		if bytes.Compare(iter.Key(), startKey) < 0 {
+			log.L().Fatal("key out of order", zap.Binary("current", iter.Key()), zap.Binary("fist", startKey))
+		}
```
kennytm, line 438: i doubt if this check is necessary...

lightning/backend/local.go
```diff
@@ -0,0 +1,1159 @@
...
+	return leaderPeerMetas, nil
+}
+
+func (local *local) Ingest(ctx context.Context, meta *sst.SSTMeta, region *split.RegionInfo) (*sst.IngestResponse, error) {
+
+	select {
+	case <-ctx.Done():
+		return nil, nil
+	default:
+	}
```
kennytm, line 503: i don't think we should do this. we should let `Ingest()` naturally fail with `context.Canceled`.

lightning/backend/local.go
```diff
@@ -0,0 +1,1159 @@
...
+
+	iter := engineFile.db.NewIter(nil)
+	defer iter.Close()
+
+	var startKey, endKey []byte
+	if iter.First() {
+		startKey = append([]byte{}, iter.Key()...)
+	} else {
+		return nil, errors.Errorf("could not find first pair, this shouldn't happen")
```
kennytm, line 543:
```suggestion
		return nil, errors.New("could not find first pair, this shouldn't happen")
```

lightning/backend/local.go
```diff
@@ -0,0 +1,1159 @@
...
+	}
+	if iter.Last() {
+		endKey = append([]byte{}, iter.Key()...)
+	} else {
+		return nil, errors.Errorf("could not find last pair, this shouldn't happen")
```
kennytm, line 548:
```suggestion
		return nil, errors.New("could not find last pair, this shouldn't happen")
```

lightning/backend/local.go
```diff
@@ -0,0 +1,1159 @@
...
+
+func (local *local) writeAndIngestByRange(
+	ctx context.Context,
+	engineFile *LocalFile,
+	start, end []byte) error {
+	select {
+	case <-ctx.Done():
+		return errors.New("context is cancel by other reason")
+	default:
+	}
```
kennytm, line 769: remove this

lightning/backend/local.go
```diff
@@ -0,0 +1,1159 @@
...
+	pairEnd := append([]byte{}, iter.Key()...)
+
+	var regions []*split.RegionInfo
+	var err error
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+WriteAndIngest:
+	for retry := 0; retry < maxRetryTimes; retry++ {
+		if retry != 0 {
+			time.Sleep(time.Second)
```
kennytm, line 798: This sleep should be `select`ed

```suggestion
			select {
			case <-time.After(time.Second):
			case <-ctx.Done():
				return ctx.Err()
			}
```

cmd/tidb-lightning-ctl/main.go
```diff
@@ -219,15 +213,37 @@ func checkpointErrorDestroy(ctx context.Context, cfg *config.Config, tls *common
...
+		for _, table := range targetTables {
+			for engineID := table.MinEngineID; engineID <= table.MaxEngineID; engineID++ {
+				fmt.Fprintln(os.Stderr, "Closing and cleaning up engine:", table.TableName, engineID)
+				_, eID := kv.MakeUUID(table.TableName, engineID)
+				file := kv.LocalFile{Uuid: eID}
+				err := file.Cleanup(cfg.TikvImporter.SortedKVDir)
+				if err != nil {
+					fmt.Fprintln(os.Stderr, "* Encountered error while cleanup engine:", err)
+					lastErr = err
+				}
```
kennytm, line 246: why can't this use the common API (`UnsafeCloseEngine` and `Cleanup`)

lightning/backend/local.go
glorv: When lightning exits before finished (e.g. by user manually terminates), backend will be closed, but currently data in each engine hasn't been imported. So we can't deleted all the data file if checkpoint is enable.

cmd/tidb-lightning-ctl/main.go
3pointer: For importer backend, engine was stored in importer memory, we can retrieve it from alive importer process. but in local backend, if we want to use common API, we need lightning process alive or persistent engine map.  both way seems unnecessary if we only need to cleanup specify engine directory.
so we didn't choose to use common API.

kennytm: okay please include the rationale as comment in the code.

lightning/backend/local.go
kennytm: @glorv the question was about the case when the checkpoint is *disabled*, where the other condition `common.IsEmptyDir(local.localStoreDir)` was triggered.

lightning/backend/local.go
```diff
@@ -0,0 +1,1136 @@
...
+
+// Cleanup remove meta and db files
+func (e *LocalFile) Cleanup(dataDir string) error {
+	metaPath := filepath.Join(dataDir, e.Uuid.String()+engineMetaFileSuffix)
+	if _, err := os.Stat(metaPath); err == nil {
+		err = os.Remove(metaPath)
+		if err != nil {
+			return err
+		}
+	}
```
kennytm, line 102:
```suggestion
	if err := os.Remove(metaPath); err != nil && !os.IsNotExist(err) {
		return errors.Trace(err)
	}
```

lightning/backend/local.go
glorv: The dir is empty if lightning is finished successfully, thus all engine data in this directory is clean up, then this dir will be empty. We created this dir in `NewLocalBackend`, so we should remove here.

tests/_utils/run_lightning_local
```diff
@@ -0,0 +1,33 @@
...
+    --ca "$TEST_DIR/tls/ca.pem" \
+    --cert "$TEST_DIR/tls/lightning.pem" \
+    --key "$TEST_DIR/tls/lightning.key" \
+    --log-file "$TEST_DIR/lightning.log" \
+    --tidb-port 4000 \
+    --pd-urls '127.0.0.1:2379' \
+    --config "tests/$TEST_NAME/config.toml" \
+    -d "tests/$TEST_NAME/data" \
+    --backend "local" \
+    --sorted-kv-dir "tests/$TEST_NAME/sorted" \
```
kennytm, line 30: no you should not write temporary data into the source directory. just keep using `$TEST_DIR/.../`.

tests/concurrent-restore/run.sh
```diff
@@ -45,4 +45,4 @@ done
 
 # check tikv_gc_life_time is recovered to the original value
 run_sql 'select VARIABLE_VALUE from mysql.tidb where VARIABLE_NAME = "tikv_gc_life_time"';
-check_contains "VARIABLE_VALUE: $ORIGINAL_TIKV_GC_LIFE_TIME"
+check_contains "VARIABLE_VALUE: $ORIGINAL_TIKV_GC_LIFE_TIME"
```
kennytm, line 48: nit pick: trailing new lines accidentally removed.

lightning/restore/restore.go
```diff
@@ -1570,16 +1614,20 @@ func (cr *chunkRestore) deliverLoop(
...
+		// TODO: for local backend, we may save checkpoint more frequently, e.g. after writen
+		//10GB kv pairs to data engine, we can do a flush for both data & index engine, then we
+		// can safely update current checkpoint.
 	}
 
 	return
 }
 
-func (cr *chunkRestore) saveCheckpoint(t *TableRestore, engineID int32, rc *RestoreController) {
+func SaveCheckpoint(rc *RestoreController, t *TableRestore, engineID int32, chunk *ChunkCheckpoint) {
```
kennytm, line 1630: why making this function public?

lightning/backend/local.go
```diff
@@ -0,0 +1,1142 @@
...
+	if err != nil {
+		return err
+	}
+	metaPath := filepath.Join(local.localStoreDir, engine.Uuid.String()+engineMetaFileSuffix)
+	f, err := os.Create(metaPath)
+	if err != nil {
+		return err
+	}
+	_, err = f.Write(jsonBytes)
+	return err
```
kennytm, line 279:
```suggestion
	return errors.Trace(ioutil.WriteFile(metaPath, jsonBytes, 0666))
```

lightning/config/config.go
```diff
@@ -386,17 +396,26 @@ func (cfg *Config) Adjust() error {
...
+			cfg.TikvImporter.RangeConcurrency = 32
+		}
 	default:
 		return errors.Errorf("invalid config: unsupported `tikv-importer.backend` (%s)", cfg.TikvImporter.Backend)
 	}
 
+	if cfg.TikvImporter.Backend == BackendLocal {
+		if len(cfg.TikvImporter.SortedKVDir) == 0 {
+			return errors.Errorf("tikv-importer.sorted-kv-dir must not be empty!")
+		}
```
kennytm, line 416: include a check somewhere that `RegionSplitSize` must not be zero or negative.

lightning/backend/local.go
```diff
@@ -0,0 +1,1142 @@
...
+				return nil, err
+			}
+			_, endIndexID, endValues, err := tablecodec.DecodeIndexKeyPrefix(lastKeyOfIndex)
+			if err != nil {
+				return nil, err
+			}
+
+			if startIndexID != endIndexID {
+				// this shouldn't happen
+				log.L().Error("index ID not match",
```
kennytm, line 585:
```suggestion
				log.L().DPanic("index ID not match",
```

lightning/backend/local.go
```diff
@@ -0,0 +1,1142 @@
...
+		}
+		endHandle, err := tablecodec.DecodeRowKey(endKey)
+		if err != nil {
+			return nil, err
+		}
+		step := (endHandle - startHandle) / n
+		index := int64(0)
+		var sKey, eKey []byte
+
+		log.L().Info("data engine", zap.Int64("step", step),
```
kennytm, line 625: use a more specific description than "data engine".

lightning/backend/localhelper.go
```diff
@@ -0,0 +1,286 @@
...
+	"go.uber.org/zap"
+
+	"github.com/pingcap/tidb-lightning/lightning/log"
+)
+
+const SplitRetryTimes = 32
+
+// TODO remove this file and use br internal functions
+// This File include region split & scatter operation just like br.
+// we can simply call br function, but we need to change some function signature of br
```
kennytm, line 26: please file an issue on BR to track this.

3pointer:
test failed because we change saveCh to a buffer channel.

glorv:
In lightning currently implement, we will update chunk checkpoint after each `writeRows`.But in importer implement, we write to rocksdb with WAL disabled. So if importer exited there maybe data loss and can't be recovery. With local backend, if lightning exit, it's like both lighting and importer are exited. So we choose to update chunk checkpoint after engine closed, to make sure there won't be data loss. 
@kennytm  Am I right or I have missed something? Are there other approaches can use to make checkpoint more precise?

3pointer:
@kennytm @july2993  PTAL
1. coverage will increased after we use integrate br to split & scatter. 
2. in this PR, we introduce pd package into lightning, so maybe we should use multiple branch from now.

lightning/backend/local.go
```diff
@@ -0,0 +1,1149 @@
...
+		requests = append(requests, req)
+	}
+
+	bytesBuf := newBytesBuffer()
+	defer bytesBuf.destroy()
+	pairs := make([]*sst.Pair, 0, local.batchWriteKVPairs)
+	count := 0
+	size := int64(0)
+	totalCount := 0
+	for iter.First(); iter.Valid(); iter.Next() {
```
july2993, line 421: potential issue and opt:

since we disable the compaction of pebble, all the files are in L0 and may overlap, so at the worst case we will scan the whole DB data here, depend on the order the kvs are written. it's poor to scan the whole db data for every region here.

kennytm:
> in this PR, we introduce pd package into lightning, so maybe we should use multiple branch from now.

How does this affect 3.1 / 3.0 target clusters (which local backend won't support anyway)?

lightning/backend/local.go
```diff
@@ -0,0 +1,1149 @@
...
+	enableCheckpoint bool,
+) (Backend, error) {
+	pdCli, err := pd.NewClient([]string{pdAddr}, tls.ToPDSecurityOption())
+	if err != nil {
+		return MakeBackend(nil), errors.Annotate(err, "construct pd client failed")
+	}
+	splitCli := split.NewSplitClient(pdCli, tls.TLSConfig())
+
+	shouldCreate := true
+	if enableCheckpoint {
```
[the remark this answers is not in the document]
glorv: if `!enabelCheckpoint` we should return an error if the local storage directory already exists because it may contains invalid data.

glorv:
> > in this PR, we introduce pd package into lightning, so maybe we should use multiple branch from now.
> 
> How does this affect 3.1 / 3.0 target clusters (which local backend won't support anyway)?

local mode depends on some new interface like https://github.com/tikv/tikv/pull/7459 which only available in 4.0 or higher, so we do only support 4.0 or later.  
By design, local mode is only available for 4.0 or later.

kennytm:
> local mode depends on some new interface like tikv/tikv#7459 which only available in 4.0 or higher, so we do only support 4.0 or later.
> By design, local mode is only available for 4.0 or later.

Yes, I mean, since 3.x can't use local mode anyway, why do we need multiple branches?

3pointer:
> > local mode depends on some new interface like [tikv/tikv#7459](https://github.com/tikv/tikv/pull/7459) which only available in 4.0 or higher, so we do only support 4.0 or later.
> > By design, local mode is only available for 4.0 or later.
> 
> Yes, I mean, since 3.x can't use local mode anyway, why do we need multiple branches?

If so, we do not need multiple branches, only version check is needed when use local backend on 3.x TiDB cluster.

kennytm:
Yes --check-requirements for local backend already checked 4.0.0+

PR #330, opened 2020-06-10
Fix test cases on release-3.0

kennytm:

Update the test cases so that it can be run on release-3.0 again.

Avoid running 'local' backend tests when the cluster is below 4.0.0

### Check List 

Tests 

 - Integration test

Side effects

Related changes


Jenkinsfile
```diff
@@ -0,0 +1,238 @@
+def TIKV_BRANCH = "master"
```
glorv, line 1: Why we need to put this jenkinsfile in lightning repo? Currently only `binlog` use similar method to maintain ci pipeline scripts are we going to put ci scripts to other repo too? And we have a repo https://github.com/pingcap/sre that used to maintain all sre scripts(include ci scripts).
In my experience in maintaining CI, I think it's not a good idea to put this script here. In most time, these scripts are managed by QA or EE teams, only few changes are related to test logic changes, and the CI maintainers should take responsibility to manage this instead of developers

glorv:
rest LGTM

Jenkinsfile
kennytm: yes we could put it in SRE. 

wdyt @3pointer

kennytm: pingcap/SRE#339. 

Will delete if that PR is merged.

kennytm:
```
[2020-06-16T08:05:03.935Z] build github.com/pingcap/tidb-lightning/cmd/tidb-lightning: cannot load golang.org/x/exp/rand: golang.org/x/exp@v0.0.0-20200513190911-00229845015e: Get https://goproxy.onetool.net/golang.org/x/exp/@v/v0.0.0-20200513190911-00229845015e.zip: dial tcp: lookup goproxy.onetool.net on 10.233.0.10:53: no such host
```

kennytm:
PTAL @3pointer

PR #272, opened 2020-02-26
lightning: split large csv file if possible

XuHuaiyu:
### What problem does this PR solve? 
Split a large csv file if possible.

### What is changed and how it works?
Split a large csv file into multiple regions, each region is about 256MB.
Then the multiple regions can be processed concurrently, and the import phase should be accelerated.

Before this commit, we need to take 40m15s to restore a 33G a tpcc csv file.
After this commit, we only need to take 8m15s to restore the same file.

### Check List 

Tests 

 - Unit test

 - Manual test (add detailed scripts or steps below)
case 1:
1. import data using lightning **with** this feature
select count(*) from t;  --> 60,000,000

2. import data using lightning **without** this feature
select count(*) from tbl_name; --> 60,000,000

case 2:
``` sql
mysql> select c_since from bmsql_customer where c_w_id = 18 and c_d_id=1 and c_id=1;
+---------------------+
| c_since             |
+---------------------+
| 2020-02-26 20:06:00 | -- select the first line of csv file, result is correct
+---------------------+
1 row in set (2.26 sec)

mysql> select c_since from bmsql_customer where c_w_id = 2000 and c_d_id=10 and c_id=3000;
+---------------------+
| c_since             |
+---------------------+
| 2020-02-26 20:42:29 | -- select the last line of csv file, result is correct
+---------------------+
```

Side effects

N/A

Related changes

N/A

PR #331, opened 2020-06-11
config: update example config file

glorv:
Add descriptions for new fields

In #272 and #326 we introduced some new config fields but didn't add them to the example config toml.

BTW, as the fields `mydumper.strict-format` and `mydumper.max-region-size` only affact csv files, shall we move them to `mydumper.csv`?

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code


tidb-lightning.toml
```diff
@@ -79,6 +79,16 @@ addr = "127.0.0.1:8287"
 #  - ignore: keep the old record and ignore the new record (i.e. insert rows using "INSERT IGNORE INTO")
 #  - error: stop Lightning and report an error (i.e. insert rows using "INSERT INTO")
 #on-duplicate = "replace"
+# split regions into this size according to the importing data. This should be set to the same as tikv region size
+# to avoid tikv further split region. The default value is 96M.
```
kennytm, line 83:
```suggestion
# Maximum KV size of SST files produced in the 'local' backend. This should be the same as
# the TiKV region size to avoid further region splitting. The default value is 96 MiB.
```

tidb-lightning.toml
```diff
@@ -113,6 +123,13 @@ no-schema=false
 # different objects. Currently only affects [[routes]].
 case-sensitive = false
 
+# if strict-format is ture, lightning will use '\r\n' to determine the end of each line. Make sure your data
+# doesn't contain '\r\n' if strict-format is enabled, or csv parser may parse incorrect result.
```
kennytm, line 127:
```suggestion
# if strict-format is ture, lightning will use '\r' and '\n' to determine the end of each line. Make sure your data
# doesn't contain '\r' or '\n' if strict-format is enabled, or csv parser may parse incorrect result.
```

tidb-lightning.toml
```diff
@@ -113,6 +123,13 @@ no-schema=false
...
+strict-format = false
+# if strict-format is true, large csv files will be split to multi regions, each region size is `max-region-size`,
+# thus lightning can restore these regions in parallel. The defalut values is 256M.
```
kennytm, line 130:
```suggestion
# if strict-format is true, large CSV files will be split to multiple chunks, which Lightning
# will restore in parallel. The size of each chunk is `max-region-size`, where the default is 256 MiB.
```

PR #334, opened 2020-06-17
optimize parse csv and local backend write tikv

glorv:
optimize local backend write to tikv and csv parser.

- Reuse the object in `local.WriteToTikv`. perf flame graph show that 1/4 of the cpus are used in runtime.newobject, so reused the objects created after the first round loop can eliminate these costs.
- Currently, we use `bytes.IndexAny` to find the stop position when read csv line. This method will always create an `asciiSet` for every call. By split the original function into two function for ASCII and UTF-8 strings and pre-generate the asciiSet, we can avoid this cost and also make this call faster.

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Side effects

Related changes


glorv:
@kennytm @3pointer PTAL

lightning/mydump/bytes.go
```diff
@@ -0,0 +1,69 @@
...
+    for i, c := range s {
+        if as.contains(c) {
+            return i
+        }
+    }
+    return -1
+
+}
+
+func IndexAnyUtf8(s []byte, chars string) int {
```
kennytm, line 53: we don't need to care about UTF-8. the current CSV restriction is that the separator must be one *byte* long and the delimiter must be zero or one *byte* long, which leaves no room for non-ASCII characters.

glorv: The implement is consistent with both  ascii and utf-8 string and with automatically use either `IndexAnyAscii` or `IndexAnyUtf8`, so we can just left this function so it can be used if later we support utf8 delimiters

kennytm: the result will be wrong if the buffer is cut like:

```
[20 20 20 20 20 20 20 20 C3]
[A4 20 20 20 20 20 20 20 20]
```

if we search for `C3 A4` (ä) we will not found anything using `IndexAnyUtf8`.

glorv: yes, based on current implement, it's not easy to handle this, so I removed `IndexAnyUtf8` related code

lightning/backend/local.go
```diff
@@ -257,7 +257,7 @@ func (local *local) ShouldPostProcess() bool {
 
 func (local *local) openEngineDB(engineUUID uuid.UUID, readOnly bool) (*pebble.DB, error) {
 	opt := &pebble.Options{
-		MemTableSize:             128 << 20,
+		MemTableSize:             512 << 20,
```
3pointer, line 260: should we give this value by current memory size?

glorv: In my benchmark for 1TB data import, this change doesn't impact much for memory usage(it will cause memory usage increase faster, but peak memory isn't increase much), and do have some performance improve. So I think this is ok. 
And if we want to set it based on system memory size, we should do more benchmarks to see the impact in different configs.

3pointer: OK, maybe leave a TODO

3pointer:
LGTM

PR #338, opened 2020-06-24
optimize encoder and adjust some config

glorv:
optimize the encoder preformance

- Apply https://github.com/pingcap/tidb/pull/18102 this optimization to make chunk encoder run faster.
- adjust some config's default value to reduce the memory usage:
  - set default `range-concurrency` to 16. 16 is big enough to take full use of a 10Gb(1.25GB) bandwidth network. So it will fit the physical environment in most case.
  - set default `send-kv-pairs` to 32768. This can reduce lightning peak memory usage from 19G to 10G without performance loss.
  - remove the misleading comment for `sorted-kv-dir`.
  - Decrease the `batch-size` from default 107374182400(100G) to 10G for local backend. In my benchmark can make about 6% performance increase

Benchmark result:
Import 14k warehouse (1007G) data with 1 lightning in our idc machines. With following config:
```
[lightning]
index-concurrency = 4
table-concurrency = 4

[checkpoint]
enable = false

[tikv-importer]
backend = "local"
range-concurrency = 16
send-kv-pairs = 32768

[mydumper]
batch-size = 10737418240
```

with current master branch it takes **2h27min**. With this pr it takes **2h6min**.

Import 14k warehouse tpcc data with importer backend takes **2h9min**.

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Side effects

Related changes


tidb-lightning.toml
```diff
@@ -83,11 +83,12 @@ addr = "127.0.0.1:8287"
...
 #region-split-size = 100_663_296
 # write key-values pairs to tikv batch size
-#send-kv-pairs = 100000
-# local storage directory used in "local" backend. Put this direcory in a different disk from `data-source-dir`
-# can make lightning run faster.
+#send-kv-pairs = 32768
+# local storage directory used in "local" backend.
 #sorted-kv-dir = ""
 # range-concurrency controls the maximum ingest concurrently while writing to tikv, It can affect the network traffic.
+# this default config can make full use of a 10Gib bandwidth network, if the newwork bandwidth is higher, you can increate
```
kennytm, line 90:
```suggestion
# this default config can make full use of a 10Gib bandwidth network, if the network bandwidth is higher, you can increase
```

tidb-lightning.toml
```diff
@@ -83,11 +83,12 @@ addr = "127.0.0.1:8287"
...
+# this to gain better performance. Bigger config value will also increate the memory usage slightly.
```
kennytm, line 91:
```suggestion
# this to gain better performance. Larger value will also increase the memory usage slightly.
```

lightning/backend/session.go
```diff
@@ -94,9 +101,10 @@ func (t *transaction) SetAssertion(kv.Key, kv.AssertionType) {}
 // transaction type and provides the session variables to the TiDB library
 // optimized for Lightning.
 type session struct {
-	sessionctx.Context
```
kennytm, line 97: keep this, and remove all methods not called, so that we'll know what is the proper implementation when they are actually called.

glorv: We have called `Context.SetValue` to reuse the `CommonAddRecordCtx`.  If we keep this, the Context is nil, thus cause panic with nil pointer.

lightning/config/config.go
```diff
@@ -504,7 +504,7 @@ func (cfg *Config) Adjust() error {
 
 	// handle mydumper
 	if cfg.Mydumper.BatchSize <= 0 {
-		cfg.Mydumper.BatchSize = 100 * _G
+		cfg.Mydumper.BatchSize = 10 * _G
```
kennytm, line 507: keep the default as 100 GB when the backend is "importer".

glorv: I have test 1T data with importer backend with 10G as the default config, seems there is no obvious impact.

PR #343, opened 2020-07-06
server: check open file ulimit for local backend

### LINKED ISSUE #342: local backend error: too many open files
## Bug Report

When importing massive data (2.8T, 40k warehouses tpcc data), tidb-lightning will raise error "too many open files" constantly without quit.

Check the limit files before importing

3. What did you see instead?
Error "too many open files" all the time

4. Versions of the cluster

    - TiDB-Lightning version (run `tidb-lightning -V`):

        ```
        [root@localhost tidb-lightning]# ./bin/tidb-lightning -V
        Release Version: v3.0.16-2-g4af19d5
        Git Commit Hash: 4af19d5d44837d6c6fe3ba3ad191df354e310ad3
        Git Branch: master
        UTC Build Time: 2020-07-02 10:23:10
        Go Version: go version go1.13.11 linux/amd64
        ```
        
    - TiKV-Importer version (run `tikv-importer -V`)

        ```
        (paste TiKV-Importer version here)
        ```
        
    - TiKV version (run `tikv-server -V`):

        ```
        (paste TiKV version here)
        ```

    - TiDB cluster version (execute `SELECT tidb_version();` in a MySQL client):

        ```
        (paste TiDB cluster version here)
        ```

    - Other interesting information (system version, hardware config, etc):

        >
        >

5. Operation logs
   - Please upload `tidb-lightning.log` for TiDB-Lightning if possible
   - Please upload `tikv-importer.log` from TiKV-Importer if possible
   - Other interesting logs
   
6. Configuration of the cluster and the task
   - `tidb-lightning.toml` for TiDB-Lightning if possible
   - `tikv-importer.toml` for TiKV-Importer if possible
   - `inventory.ini` if deployed by Ansible

7. Screenshot/exported-PDF of Grafana dashboard or metrics' graph in Prometheus for TiDB-Lightning if possible

glorv:
Check process max open file limit for local backend

In local backend mode, pebble need to write a lot of L0 sst files. So If data source is fairly big, the max_open_file limit may not be enough. So we check the `RLIMIT_NOFILE` base on source file size, and try to update it if not big enough.

close #342 

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Side effects

Related changes

### DIFF
--- a/cmd/tidb-lightning/main.go
+++ b/cmd/tidb-lightning/main.go
@@ -17,11 +17,8 @@ import (
 	"fmt"
 	"os"
 	"os/signal"
-	"path/filepath"
 	"runtime/debug"
 	"syscall"
-
-	"github.com/pingcap/errors"
 
 	"github.com/pingcap/tidb-lightning/lightning"
 	"github.com/pingcap/tidb-lightning/lightning/config"
@@ -32,12 +29,6 @@ import (
 func main() {
 	cfg := config.Must(config.LoadGlobalConfig(os.Args[1:], nil))
 	fmt.Fprintf(os.Stdout, "Verbose debug logs will be written to %s.\n\n", cfg.App.Config.File)
-	err := checkSystemRequirement(cfg)
-	if err != nil {
-		log.L().Error("check system requirements failed", zap.Error(err))
-		fmt.Fprintln(os.Stderr, "check system requirements failed:", err)
-		return
-	}
 
 	app := lightning.New(cfg)
 
@@ -75,7 +66,7 @@ func main() {
 		}
 	}
 
-	err = app.GoServe()
+	err := app.GoServe()
 	if err != nil {
 		logger.Error("failed to start HTTP server", zap.Error(err))
 		fmt.Fprintln(os.Stderr, "failed to start HTTP server:", err)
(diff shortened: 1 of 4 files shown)


cmd/tidb-lightning/main.go
```diff
@@ -96,3 +105,47 @@ func main() {
 		os.Exit(1)
 	}
 }
+
+func checkSystemRequirement(cfg *config.GlobalConfig) error {
```
kennytm, line 109: (1) This should be checked using the task config before every task starts
(2) This should be ignorable using `--check-requirements=false`

cmd/tidb-lightning/main.go
```diff
@@ -96,3 +105,47 @@ func main() {
...
+		}
+		if rLimit.Max < estimateMaxFiles {
+			// If the process is not started by privileged user, this will fail.
+			rLimit.Max = estimateMaxFiles
+		}
+		prevLimit := rLimit.Cur
+		rLimit.Cur = estimateMaxFiles
+		err = syscall.Setrlimit(syscall.RLIMIT_NOFILE, &rLimit)
+		if err != nil {
+			return errors.Wrap(err, fmt.Sprintf("the maximum number of open file descriptors is too small, got %d, expect greater or equal to %d", prevLimit, estimateMaxFiles))
```
kennytm, line 146:
```suggestion
			return errors.Annotatef(err, "the maximum number of open file descriptors is too small, got %d, expect greater or equal to %d", prevLimit, estimateMaxFiles)
```

cmd/tidb-lightning/main.go
```diff
@@ -96,3 +105,47 @@ func main() {
...
+			}
+			return err
+		})
+		if err != nil {
+			return errors.Trace(err)
+		}
+
+		// we estimate each sst as 50MB, this is a fairly coarse predict, the actual fd needed is
+		// depend on chunk&import concurrency and each data&index key-value size
+		estimateMaxFiles := totalSize / (50 * 1024 * 1024)
```
kennytm, line 129: how this "50 MiB" is determined? 

also, for a 10 TB data set this means Pebble will open up to 200000 files simultaneously. This sounds more like a bug to me (that the engines are not closed quickly enough).

glorv: In the tpcc benchmaerk, with a memtable size of 512MB, the generated L0 sst file sizes ranges from 40~450MB, so I'm not sure how to estimate the avg size. And In the common case, L0 sst files will be compressed to L1~Ln level ssts, thus total files should be very large, but we prevent the compression phase in favor of better performance thus the generated files numbers will be quite big. Because index engine can't be split, so if a table is huge, and there are many indices, then maybe we will generate thousands of L0 sst files.

kennytm: can we use the largest N total size of each table (N = index concurrency) instead of the entire data source?

from the pebble source code it seems 50 MiB = [25](https://github.com/cockroachdb/pebble/blob/a452ae2c916a74a035d264e42c9b761061a988b5/compaction.go#L41) × [2 MiB](https://github.com/cockroachdb/pebble/blob/a452ae2c916a74a035d264e42c9b761061a988b5/options.go#L225), so maybe we could increase the TargetFileSize? 🤔 In TiKV Importer [this is set](https://github.com/tikv/importer/blob/9531815ad95fac8dc6febcf0fb8cd76d01662df3/src/import/engine.rs#L385) to [1 GiB](https://github.com/tikv/importer/blob/9531815ad95fac8dc6febcf0fb8cd76d01662df3/etc/tikv-importer.toml#L23).

glorv: Seems the L0 sst file size is depend on the memory table size and the conf `MaxOpenFiles` is used for compaction. After explict set `MaxOpenFiles` option to 1GB, the table `order_line` index engine still generates a lots of sst files with size of 54MB. And almost each engine generates sst files with different size.  Maybe we should use `MemoryTableSize` as each sst file's size, as memory table accounts for the raw size of the kv-pairs.
So we can estimate by:  {top N table size} / {MemoryTableSize} as the needed fds

kennytm: Okay.

lightning/lightning_test.go
```diff
@@ -399,3 +401,63 @@ func (s *lightningServerSuite) TestHTTPAPIOutsideServerMode(c *C) {
...
+				},
+				{
+					TotalSize: 150_000 << 20, //200_000MB
+				},
+			},
+		},
+		{
+			Tables: []*mydump.MDTableMeta{
+				{
+					TotalSize: 150_800 << 20, //150_288MB
```
kennytm, line 424: not sure what these comments mean but the numbers looks way off.

kennytm:
Blocked by #348.

PR #345, opened 2020-07-07
log: fix log file path

glorv:
fix lightning always set log file path to the default path

Only set log path to default path if neither command line params nor config file set it 

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Side effects

Related changes

### DIFF
--- a/lightning/config/global.go
+++ b/lightning/config/global.go
@@ -197,7 +197,10 @@ func LoadGlobalConfig(args []string, extraFlags func(*flag.FlagSet)) (*GlobalCon
 	if *logFilePath != "" {
 		cfg.App.Config.File = *logFilePath
 	}
-	if cfg.App.Config.File == "" {
+	// "-" is a special config for log to stdout
+	if cfg.App.Config.File == "-" {
+		cfg.App.Config.File = ""
+	} else if cfg.App.Config.File == "" {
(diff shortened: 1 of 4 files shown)


lightning/config/global.go
```diff
@@ -197,6 +197,9 @@ func LoadGlobalConfig(args []string, extraFlags func(*flag.FlagSet)) (*GlobalCon
 	if *logFilePath != "" {
 		cfg.App.Config.File = *logFilePath
 	}
+	if cfg.App.Config.File == "" {
```
3pointer, line 200:
```suggestion
	else {
```

glorv: This else is not ok, because toml config may set the log config path

glorv:
> there should still be a way to explicitly log to terminal

Then I think we should leave the default to stdout

kennytm:
We could make `--log-file "-"` to mean stdout, or copy BR's `$BR_LOG_TO_TERM` env var

glorv:
> We could make `--log-file "-"` to mean stdout, or copy BR's `$BR_LOG_TO_TERM` env var

So shall we need to update docs for this special log config. Seems br special env `$BR_LOG_TO_TERM` is not mensioned in docs

PR #346, opened 2020-07-07
Arbitrary file import

### LINKED ISSUE #344: Support load from file without formatted file name
## Feature Request

Currently, we just support pass a folder with Mydumper-formatted files. This is in someway too "heavy", and some of other tools like `sqlldr` can load arbitrary file directly with command-line arguments.

Support importing arbitrary file will make life easier when we want to load some "incremental" csv data using TiDB backend.

add some flags:
- `-from-file` specifies a set of file to load
- `-to-table` specifies which table to load into
- `-to-db` specifies which database to load into

We can rename those files directly and then load them with our "happy path", but renaming files isn't easy sometimes...

When `-from-file` presents, we can load file and schema info from `-from-file`, `-to-table`, `-to-db` directly (omit the MyDumper folder), and `-no-schema` implies.

Then everything go on as normal.

YuJuncen:
close #344 

Before RFC 5 is implemented, use this PR at risk.

We make a demo of `-from-file`, as a practice of contributing to lightning. If there is better solution, fell free to close this.

### Check List 

Tests 

 - Integration test


glorv:
I think we can close this in favor of #366  @YuJuncen @kennytm

PR #347, opened 2020-07-07
backend: fix local backend may divide by 0

glorv:
fix a divide by 0 bug in local backend when split index engine ranges

If the index engine total size is bigger than RegionSplitSize but small than indexCount * RegionSplitSize, thus each index range count will be 0, thus split kv-pairs by range will encounter divide by 0 panic.

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Side effects

Related changes


glorv:
@3pointer  PTAL

PR #352, opened 2020-07-17
Fix verbose log message for shell

nullnotnil:

When I run a new tidb-lightning task interactively, it generates a log file for verbose output. In my shell I double-click to pickup that file so I can copy it. Unfortunately it has a trailing `.` which is annoying enough to me that you are seeing this PR :-)

Message text change

### Check List 

 - No code

Side effects

 - None

Related changes

 - None


nullnotnil:
CLA signed.

PR #356, opened 2020-07-22
local: fix batch split retry alway failed error

glorv:
When `BatchSplitRegions` return an error such as `EpochNotMatch`, current retry policy will retry with the same region info , which will sure to be failed again. Thus, upto 32 times retry won't have any effect.

If we encounter an error when `BatchSplitRegions` we should retry the whole function because the regions are probably changed.
Return the latest error if all retry are failed.

I tested this changed in a test cluster use 1k warehouse tpcc data. When set table/index to 8 or higher. Before this change this is a long tail import phase and the total time often nearly 30min.  After this change, the time drop to about 13min, which is the average speed.

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Side effects

Related changes

### DIFF
--- a/cmd/tidb-lightning/main.go
+++ b/cmd/tidb-lightning/main.go
@@ -28,7 +28,7 @@ import (
 
 func main() {
 	cfg := config.Must(config.LoadGlobalConfig(os.Args[1:], nil))
-	fmt.Fprintf(os.Stdout, "Verbose debug logs will be written to %s.\n\n", cfg.App.Config.File)
+	fmt.Fprintf(os.Stdout, "Verbose debug logs will be written to %s\n\n", cfg.App.Config.File)
 
 	app := lightning.New(cfg)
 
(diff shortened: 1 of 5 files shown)


kennytm:
/run-all-tests tidb=release-4.0 tikv=release-4.0 pd=release-4.0 importer=release-4.0

3pointer:
/run-all-tests tidb=release-4.0 tikv=release-4.0 pd=release-4.0 importer=release-4.0

glorv:
@kennytm  PTAL again. I have update the logic to only retry the failed keys. But I'm not sure which one is better

kennytm:
/run-all-tests tidb=release-4.0 tikv=release-4.0 pd=release-4.0 importer=release-4.0

glorv:
Similary logic in br has the same problem, if this is ok, we can fix that in the same way. See: https://github.com/pingcap/br/blob/master/pkg/restore/split.go#L98-L144

PR #360, opened 2020-07-29
mydump: add a warning log on large file with CSV header

kennytm:

Auto-splitting requires (1) `strict-format = true` and (2) `csv.header = false`. Since `csv.header = true` by default, it would be confusing that setting `strict-format = true` alone has no effect.

Add a warning if a large CSV file cannot be split only because of `csv.header`.

### Check List 

Tests 

 - No code

Side effects

Related changes

### DIFF
--- a/lightning/mydump/region.go
+++ b/lightning/mydump/region.go
@@ -163,18 +163,25 @@ func MakeTableRegions(
 		}
 		// If a csv file is overlarge, we need to split it into mutiple regions.
 		// Note: We can only split a csv file whose format is strict and header is empty.
-		if isCsvFile && dataFileSize > cfg.Mydumper.MaxRegionSize && cfg.Mydumper.StrictFormat && !cfg.Mydumper.CSV.Header {
-			var (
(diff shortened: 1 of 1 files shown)


glorv:
I think we can still split large csv files even if they contains header line?

kennytm:
only if we read the header before splitting. this changes the scanning logic considerably.

glorv:
LGTM. BTW, I think it's strage for user that we can't split large csv because of head line? As a use, I will think, why not just skip the first line and split the rest

kennytm:
@glorv the problem is we can't just skip the first line, since the CSV column order and `CREATE TABLE` column order are not necessarily equivalent.

overvenus:
LGTM

kennytm:
should we close this in favor of #363, or merge this first?

overvenus:
I prefer to close this if we can merge #363 before the next release (about 2020/08/21).

kennytm:
#363 has been merged.

PR #364, opened 2020-08-04
parser: fix csv parse header with empty line

glorv:
Enable csv parser to skip empty line. See: https://asktug.com/t/topic/36359/24

Csv parse will treat the first non-empty line as header instead of the first line. And will also skip empty lines.

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Side effects

Related changes

# Release notes

Empty lines in CSV are ignored.


kennytm:
The unit test entered an infinite loop.

glorv:
> The unit test entered an infinite loop.

fixed and add a unit test

PR #361, opened 2020-07-30
restore: support restore from s3 

glorv:
### What problem does this PR solve? 
Support restore data files from s3 storage
close #69

Related change in br: https://github.com/pingcap/br/pull/446

### What is changed and how it works?
Replace the direct file IO by  ExternalStorage interface and share the storage implements from br.

### Check List 

Tests 

 - Unit test
 - Integration test
 - Manual test (add detailed scripts or steps below)
 - No code

Side effects

 - Increased code complexity

Related changes

 - Need to cherry-pick to the release branch
 - Need to update the documentation
 - Need to be included in the release note

PR #366, opened 2020-08-07
restore: support file level routing

glorv:
Implement table level routing, see `Lighting RFC 5` for more detals

With now configurations `mydumper.files`, user can privide regexp pattern to match source files and conf the source relate schema/table/type either by constant string or template value expand from the regexp pattern.

NOTE: this pr doesn't implement following features:
- file sorting key. relate to the `mydumper.files.key` conf. User can still set this conf, but is currently useless.
- file sorting key transform. relate to the `mydumper.files.key-transform`. This conf is not provided.
- compression. relate to the `mydumper.files.compression`. As we don't support read compression files yet.  User can still set this conf, but is currently useless.

We should deprecate the feature `table router` and remote the `lightning.router` config fileds in the future. 

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Side effects

 - Increased code complexity
 - Increased config complexity

Related changes


kennytm:
```
[2020-08-07T10:28:24.499Z] Error: restore table `error_summary`.`a` failed: checksum mismatched remote vs local => (checksum: 10176589659643620760 vs 5152985213657377814) (total_kvs: 3 vs 2) (total_bytes:87 vs 58)
```

glorv:
> remote

This is expected. Because target table contains 1 row before restore.

lightning/config/config.go
```diff
@@ -140,17 +140,28 @@ type CSVConfig struct {
...
+	SourceDir        string           `toml:"data-source-dir" json:"data-source-dir"`
+	NoSchema         bool             `toml:"no-schema" json:"no-schema"`
+	CharacterSet     string           `toml:"character-set" json:"character-set"`
+	CSV              CSVConfig        `toml:"csv" json:"csv"`
+	CaseSensitive    bool             `toml:"case-sensitive" json:"case-sensitive"`
+	StrictFormat     bool             `toml:"strict-format" json:"strict-format"`
+	MaxRegionSize    int64            `toml:"max-region-size" json:"max-region-size"`
+	Filter           []string         `toml:"filter" json:"filter"`
+	FileRouters      []*FileRouteRule `toml:"files" json:"files"`
+	DefaultFileRules bool             `toml:"default-file-rules" json:"default-file-rules"`
```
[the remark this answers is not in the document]
glorv: yes, I considered this but forgot to update the code. It can be done in adjustConfig. But I think in some case, user need to add some rule in addition to the default rule, thus set this to true will be convenient.

I will update the logic to set the default value to false and enable it `len(FileRouters) == 0`.  And user can explicitly enable it.

lightning/mydump/router.go
```diff
@@ -0,0 +1,365 @@
...
+	case SchemaSchema:
+		return SourceTypeSchemaSchema
+	case TableSchema:
+		return SourceTypeTableSchema
+	case TypeSQL:
+		return SourceTypeSQL
+	case TypeCSV:
+		return SourceTypeCSV
+	default:
+		return SourceTypeIgnore
```
kennytm, line 54: the default case should be an error

lightning/mydump/router.go
```diff
@@ -0,0 +1,365 @@
...
+	case "xz":
+		return CompressionXZ
+	default:
+		return CompressionNone
+	}
+}
+
+var (
+	isAlphaNum = regexp.MustCompile(`^[A-Za-z0-9]+$`).MatchString
+	isAlpha    = regexp.MustCompile(`^[0-9]+$`).MatchString
```
kennytm, line 90: named "isAlpha" but checks for numbers??????

lightning/mydump/router.go
```diff
@@ -0,0 +1,365 @@
...
+var (
+	defaultFileRouteRules = []*config.FileRouteRule{
+		// ignore *-schema-view.sql,-schema-trigger.sql,-schema-post.sql files
+		{Pattern: `(?i).*(-schema-view|-schema-trigger|-schema-post)\.sql`, Type: "ignore"},
+		// db schema create file pattern, matches files like '{schema}-schema-create.sql'
+		{`(?i)^(?:[^/]*/)*([^/.]+)-schema-create\.sql`, "$1", "", SchemaSchema, "", ""},
+		// table schema create file pattern, matches files like '{schema}.{table}-schema.sql'
+		{`(?i)^(?:[^/]*/)*([^/.]+)\.(.*?)-schema\.sql`, "$1", "$2", TableSchema, "", ""},
+		// source file pattern, matches files like '{schema}.{table}.0001.{sql|csv}'
+		{`(?i)^(?:[^/]*/)*([^/.]+)\.(.*?)(\.[0-9]+)?\.(sql|csv)$`, "$1", "$2", "$4", "$3", ""},
```
kennytm, line 102: please include the field names (Pattern:, Type:, etc.)

lightning/mydump/router.go
```diff
@@ -0,0 +1,365 @@
...
+	}
+)
+
+// // RouteRule is a rule to route file path to target schema/table
+type FileRouter interface {
```
kennytm, line 107: why this is an interface, instead of a slice of `RegexRouteRule`

lightning/mydump/router.go
```diff
@@ -0,0 +1,365 @@
...
+	pattern    *regexp.Regexp
+	extractors []regexExtractor
+}
+
+func (r *RegexRouteRule) Route(path string) *RouteResult {
+	if !r.pattern.MatchString(path) {
+		return nil
+	}
+
+	indexes := r.pattern.FindStringSubmatchIndex(path)
```
kennytm, line 153:
```suggestion
	indexes := r.pattern.FindStringSubmatchIndex(path)
	if len(indexes) == 0 {
		return nil
	}
```

lightning/mydump/router.go
```diff
@@ -0,0 +1,365 @@
...
+		})
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (p *regexRuleParser) parseFieldExtractor(
```
kennytm, line 225: why do we need this function?

lightning/mydump/loader.go
```diff
@@ -244,61 +263,38 @@ func (s *mdLoaderSetup) listFiles(dir string) error {
...
+		case SourceTypeSchemaSchema:
 			s.dbSchemas = append(s.dbSchemas, info)
-		case fileTypeTableSchema:
+		case SourceTypeTableSchema:
 			s.tableSchemas = append(s.tableSchemas, info)
-		case fileTypeTableData:
+		case SourceTypeSQL, SourceTypeCSV:
 			s.tableDatas = append(s.tableDatas, info)
 		}
+
```
kennytm, line 294: please add an "unsupported" warning (and ignore the file) if `Compression` is not none

tidb-lightning.toml
```diff
@@ -153,6 +163,21 @@ backslash-escape = true
 # if a line ends with a separator, remove it.
 trim-last-separator = false
 
+# file level routing rule that map file path to schema,table,type,sort-key
+# The schema, table , type and key can be either a constant string or template strings
+# supported by go regexp. 
+#[[mydumper.files]]
+# pattern is  a regexp in Go syntax for matching the file path.
+# pattern = "(?i)^(?:[^/]*/)(?P<schema>[^/.]+)\.([^/.]+)(?:\.([0-9]+))?\.(sql|csv)$"
```
kennytm, line 171:
```suggestion
# pattern = '(?i)^(?:[^/]*/)(?P<schema>[^/.]+)\.([^/.]+)(?:\.([0-9]+))?\.(sql|csv)$'
```

use single quotes to avoid `\` escaping stuff

lightning/checkpoints/checkpoints.go
```diff
@@ -435,6 +437,10 @@ func NewMySQLCheckpointsDB(ctx context.Context, db *sql.DB, schemaName string, t
 			engine_id int unsigned NOT NULL,
 			path varchar(2048) NOT NULL,
 			offset bigint NOT NULL,
+			type int NOT NULL,
+			compression int NOT NULL,
+			sort_key varchar(256) NOT NULL,
+			size bigint NOT NULL,
```
kennytm, line 443: the *file* size doesn't look useful as part of the *chunk* checkpoint

lightning/mydump/loader.go
```diff
@@ -40,11 +38,23 @@ type MDTableMeta struct {
...
+	Path        string
+	Type        SourceType
+	Compression Compression
+	SortKey     string
+}
+
+type SourceFileMeta struct {
+	SourceMeta
+	Size int64
+}
```
kennytm, line 56: i don't think you need this type...

lightning/mydump/loader.go
```diff
@@ -146,12 +176,9 @@ func (ftype fileType) String() string {
 
 type fileInfo struct {
 	tableName filter.Table
-	path      string
-	size      int64
+	fileMeta  *SourceFileMeta
 }
```
kennytm, line 180: ... as you could keep the `size` field here

lightning/restore/restore.go
```diff
@@ -1266,13 +1277,20 @@ func newChunkRestore(
 		return nil, errors.Trace(err)
 	}
 
+	routeRes := router.Route(chunk.Key.Path)
+	if routeRes == nil || routeRes.Type == mydump.SourceTypeIgnore {
+		return nil, errors.Errorf("file '%s' is filtered by file router", chunk.Key.Path)
+	}
```
kennytm, line 1283: wait i thought the point of saving compression/type/key into the chunk checkpoint is the avoid routing again? 🤔

lightning/mydump/region.go
```diff
@@ -57,10 +57,13 @@ func (rs regionSlice) Swap(i, j int) {
...
 func (rs regionSlice) Less(i, j int) bool {
-	if rs[i].FileMeta.Path == rs[j].FileMeta.Path {
-		return rs[i].Chunk.Offset < rs[j].Chunk.Offset
+	if rs[i].FileMeta.SortKey != rs[j].FileMeta.SortKey {
+		return rs[i].FileMeta.SortKey < rs[j].FileMeta.SortKey
 	}
-	return rs[i].FileMeta.Path < rs[j].FileMeta.Path
+	if rs[i].FileMeta.Path != rs[j].FileMeta.Path {
+		return rs[i].FileMeta.Path < rs[j].FileMeta.Path
+	}
```
kennytm: the ordering doesn't look correct, as `db.z.0.sql` will be sorted before `db.a.100.sql`. 

but this function seems totally unused, so let's just delete the whole `regionSlice` type 🙂.

lightning/mydump/loader.go
glorv: Yes, I update this by using mydump.fileInfo. But this end up with exposing fileInfo to FileInfo because of using in restore

lightning/restore/restore.go
glorv: oo, I forgot to update these logic

lightning/mydump/region.go
glorv: This would happen only if `db.z.0.sql` and `db.a.100.sql` are routing to the same table. And If files with different perfix are routing to the same table, sort by sort-key should be the correct way to order them, because sort-key is designed for this

glorv: BTW, this sort is useless because at this moment, each chunk's handle alloction is already determined. Which chunk restore first will almost make no difference.

lightning/mydump/loader_test.go
```diff
@@ -11,17 +11,17 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package mydump_test
+package mydump
```
kennytm, line 14: prefer keeping the test package separate

glorv: I change to package to avoid expose `fileInfo` for test, but end up with exposing it in restore. Thus I have changed back the package

lightning/mydump/loader.go
```diff
@@ -190,48 +213,60 @@ func (s *mdLoaderSetup) setup(dir string) error {
...
 		})
+
+		// sort each table source files by sort-key
+		for _, tbMeta := range dbMeta.Tables {
+			dataFiles := tbMeta.DataFiles
+			sort.SliceStable(dataFiles, func(i, j int) bool {
+				if dataFiles[i].FileMeta.SortKey != dataFiles[j].FileMeta.SortKey {
+					return dataFiles[i].FileMeta.SortKey < dataFiles[j].FileMeta.SortKey
+				}
+				return dataFiles[i].FileMeta.Path < dataFiles[j].FileMeta.Path
```
kennytm, line 260:
```suggestion
				return dataFiles[i].FileMeta.SortKey < dataFiles[j].FileMeta.SortKey
```

the result of WalkDir is already sorted by path, comparing Path is redundant when using SliceStable

lightning/mydump/loader_test.go
```diff
@@ -20,8 +20,10 @@ import (
 	"testing"
 
 	. "github.com/pingcap/check"
+	"github.com/pingcap/tidb-tools/pkg/filter"
+	
 	"github.com/pingcap/tidb-lightning/lightning/config"
-	md "github.com/pingcap/tidb-lightning/lightning/mydump"
+	. "github.com/pingcap/tidb-lightning/lightning/mydump"
```
kennytm, line 26: um could you revert this?

lightning/mydump/router.go
glorv: I think there is no much difference to make it as a interface or an actual implement type. I prefer to use an interface type, because it's more clear for use to determine the methods this type provide.

lightning/mydump/router.go
glorv: This method is use to parse each field in `RegexRouteRule `

lightning/mydump/router.go
glorv: And I think maybe, we can Implement another `FileRouter` by just treat the pattern as file path, in this way, user needn't escape special chars in the file path, though I'm not sure how should we determine whether the input is an regex pattern or just a plain file path.

lightning/mydump/router.go
kennytm: @glorv  i mean why do we need to perform such parsing? we could just make every field a `patExpander`. the current checks make the code very long 🤔

lightning/mydump/router.go
kennytm: i see. perhaps always use `pattern = ...` for patterns and introduce `path = ...` for literal paths.

lightning/mydump/router.go
glorv: e, since the regex is kind of complex, I think we'd better do some check carefully to avoid filter out source files due to the wrong configed patterns. I will update the check to cover these patterns

glorv:
@kennytm  PTAL again.
BTW, shall we add the new config field `path` to [files] to allow user set a file path instead of regexp pattern?

kennytm:
>  BTW, shall we add the new config field `path` to `[files]` to allow user set a file path instead of regexp pattern?

do it if it's easy

---

i still think there's no point having a `constExpander`. 

---

you are also making `extractSubPatterns` way too complicated by using `strings.IndexByte`. try https://play.golang.org/p/pxPaTbLPuWH.

```go
var expandVariablePattern = regexp.MustCompile(`\$(?:\$|[\pL\p{Nd}_]+|\{[\pL\p{Nd}_]+\})`)

func VerifyExpansion(re *regexp.Regexp, template string) error {
	for _, variable := range expandVariablePattern.FindAllString(template, -1) {
		var name string
		switch {
		case variable == "$$":
			continue
		case strings.HasPrefix(variable, "${"):
			name = variable[2 : len(variable)-1]
		default:
			name = variable[1:]
		}
		if number, err := strconv.Atoi(name); err == nil {
			if number > re.NumSubexp() {
				return fmt.Errorf("sub pattern capture '%s' out of range", variable)
			}
		} else if re.SubexpIndex(name) < 0 {
			return fmt.Errorf("invalid named capture '%s'", variable)
		}
	}
	return nil
}
```

(btw a trailing `$` and malformed `${...` is fine by go standard)

glorv:
@kennytm  I updated the pattern check logic and add support for `path` config. PTAL, again 😂

BTW:
after #361  , the file path traversed by `WalkDir` will be relative path, but currently, they are all abs path. So I decide to support both abs path and relative path for `files.path` in that pr, because there are a lot of tests to modify and I don't want to do again in this pr

lightning/mydump/router.go
```diff
@@ -160,20 +163,32 @@ func (r *RegexRouter) Route(path string) *RouteResult {
...
-type regexRouterParser struct {
-	rule *RegexRouter
-	r    *config.FileRouteRule
-}
+type regexRouterParser struct{}
+
+func (p regexRouterParser) Parse(r *config.FileRouteRule) (*RegexRouter, error) {
+	rule := &RegexRouter{}
+	if r.Path != "" && r.Pattern != "" {
+		return nil, errors.New("can't set both `path` and `pattern` field")
```
kennytm, line 171:
```suggestion
		return nil, errors.New("can't set both `path` and `pattern` field in [[mydumper.files]]")
```

lightning/mydump/router.go
```diff
@@ -143,14 +144,16 @@ func NewFileRouter(cfg []*config.FileRouteRule) (FileRouter, error) {
...
 
 func (r *RegexRouter) Route(path string) *RouteResult {
-	indexes := r.pattern.FindStringSubmatchIndex(path)
-	if len(indexes) == 0 {
+	// the regexp pattern maybe not contains any sub matches
+	if !r.pattern.MatchString(path) {
 		return nil
 	}
+
+	indexes := r.pattern.FindStringSubmatchIndex(path)
```
kennytm: ???

even if `r.pattern` has no sub matches, if there is match `FindStringSubmatchIndex` will always return a non-empty array, since `$0` refers to the whole match.

lightning/mydump/router.go
```diff
@@ -262,115 +278,48 @@ func (p *regexRouterParser) parseFieldExtractor(
...
-				if c == n {
-					continue outer
-				}
-			}
-			return errors.Errorf("invalid named capture '%s', available names are: %s", c, p.rule.pattern.SubexpNames())
+func (p regexRouterParser) checkSubPatterns(pat *regexp.Regexp, t string, validateFn func(string) error) error {
+	subPats := expandVariablePattern.FindAllString(t, -1)
+	if len(subPats) == 0 {
+		// validate const patterns
+		if err := validateFn(t); err != nil {
```
kennytm, line 296: `len(subPats) > 0` does not mean `t` isn't a const pattern (`expandVariablePattern` matches `$$`). 

i suggest removing `validateFn` and let `applyFn` do the validation naturally. after all, routing is the first thing being done so delaying the validation won't cause errors being caught too late.

lightning/mydump/router.go
```diff
@@ -262,115 +278,48 @@ func (p *regexRouterParser) parseFieldExtractor(
...
-		case endIdx == 0:
-			return nil, errors.Errorf("invalid template '%s', named capture can only contain alphabets or numbers", t)
-		case endIdx < 0:
-			res = append(res, remain)
-			remain = remain[:0]
-		default:
-			res = append(res, remain[:endIdx])
-			remain = remain[endIdx:]
+		} else if !slice.AnyOf(pat.SubexpNames(), func(i int) bool {
+			// FIXME: we should use re.SubexpIndex here, but not supported in go1.13 yet
```
kennytm, line 315: 🤔 we are still using Go 1.13?!

lightning/mydump/router.go
glorv: There is a little difference. If the pattern is an invalid constant pattern for `type` I think It's better to report an error to user. But I the runtime, we will treat the pattern valid and just skip the file because maybe the file will match another pattern that return valid result.

lightning/mydump/router.go
glorv: yes, in our CI, the go version is still 1.13? The QA team recently started testing updating go version to 1.14/1.15 but  not finished yet.

kennytm: okay 😓

lightning/mydump/router.go
kennytm: no encountering an unknown file or compression type should result in an error, not skipping the file. use the `ignore` type to skip the file.

kennytm (commented):
LGTM, thank you very much!

kennytm:
PTAL @3pointer

lightning/mydump/loader_test.go
```diff
@@ -20,9 +20,11 @@ import (
 	"testing"
 
 	. "github.com/pingcap/check"
+	"github.com/pingcap/tidb-tools/pkg/filter"
```
3pointer, line 23: `filter "github.com/pingcap/tidb-tools/pkg/table-filter"  `
should import table-filter?

tidb-lightning.toml
```diff
@@ -153,6 +163,24 @@ backslash-escape = true
...
+# The schema, table , type and key can be either a constant string or template strings
+# supported by go regexp. 
+#[[mydumper.files]]
+# pattern and path determine target source files, you can use either of them but not both.
+# pattern is a regexp in Go syntax that can match one or more files in `source-dir`.
+#pattern = '(?i)^(?:[^/]*/)(?P<schema>[^/.]+)\.([^/.]+)(?:\.([0-9]+))?\.(sql|csv)$'
+# path is the target file path, both absolute file path or relative path to `mydump.source-dir` are supported.
+#path = "schema_name.table_name.00001.sql"
+# schema(database) name
+#schema = "$schema"
```
3pointer, line 176: #schema = "$1" 🤔️

lightning/mydump/loader.go
```diff
@@ -102,12 +113,26 @@ func NewMyDumpLoader(cfg *config.Config) (*MDLoader, error) {
 		f = filter.CaseInsensitive(f)
 	}
 
+	fileRouteRules := cfg.Mydumper.FileRouters
+	if cfg.Mydumper.DefaultFileRules {
+		fileRouteRules = append(fileRouteRules, defaultFileRouteRules...)
+	}
+	if len(fileRouteRules) == 0 {
+		return nil, errors.New("not file route rules. You may set 'mydumper.default-route-rules' to true or add 'mydumper.files' configs")
```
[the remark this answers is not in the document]
glorv: yes, this is redundant now

tidb-lightning.toml
glorv: `$1` is same as `$schema` here, this is an example for named capture

3pointer (approved):
LGTM

PR #369, opened 2020-08-12
backend: split and ingest region size more precise

### LINKED ISSUE #365: checksum fails easily by tikv timeout when restore big table with local backend
## Bug Report

When uses lightning to restore big table with local backend, I have seem several times in lightning logs with follows:
```
[2020/08/05 21:05:39.713 +08:00] [ERROR] [main.go:82] ["tidb lightning encountered error stack info"] [error="restore table `test`.`t` failed: compute remote checksum failed: Error 9002: TiKV server timeout"] 
```

And further more, in some benchmark, after finishing load data, when manually exec `select count(*) from table` with big table, the select may fail with:
```
mysql> select count(*) from t;
ERROR 1105 (HY000): Execution terminated due to exceeding the deadline
```

By consulting @breeswish ,  The root cause for this is that tikv process subtask in one region execeeded 1 minute, Either because the region is too large or the task waited for too long.

I thinking the primary cause of this is that  With local backend, region key range is not accurate the size of 96M, thus maybe some region is too big.

checksum should success without retry and select count(*) should return successfully

3. What did you see instead?

4. Versions of the cluster

5. Operation logs
   
6. Configuration of the cluster and the task
   

7. Screenshot/exported-PDF of Grafana dashboard or metrics' graph in Prometheus for TiDB-Lightning if possible

glorv:
try to resolve #365 by avoid writing and ingesting big regions 

make ingest region size not exceed `regionSplitSize` by:
- use sample data to calculate kv range for each region. 
 - Sample. When split range, first split the range from start-key to end-key to (total kv count) / 100 ranges. Then seek to the key that is greater or equal to each range start. If the seeked key is smaller than the range end, add the range end to sample values. If there is enough sampled values, then split the sample values to the target count and each split end with be the range end. If there are no enough sample values, we will fall back to the old naive way to split, but will use the db iterator the check empty ranges and skip them.
- Region Size. when split regions, use 3/4 of the splitRegionSize to calculate the split size. And when ingest, use 5/4 splitRegionSize as the threshold. At each write and ingest phase, write no more than this threshold byte to tikv and ingest it. If there are remian kvs, split target region and try to ingest the remain bytes. By this way, we can make sure each region size will not exceed 5/4 * splitRegionSize in bytes.

This pr also emit some useless logs to make the error and warn logs less annoying.

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Side effects

   - benchmark with 14k warehouse tpcc data shows no performance regression
 - Increased code complexity

Related changes


glorv:
sad for the code coverage decrease, the new logic for split need large scale test data

lightning/backend/local.go
```diff
@@ -365,14 +367,22 @@ func (local *local) getImportClient(ctx context.Context, peer *metapb.Peer) (sst
...
 	}
+	if bytes.Compare(startKey, start) < 0 {
+		startKey = start
+	}
 	if len(region.Region.EndKey) > 0 {
 		_, endKey, _ = codec.DecodeBytes(region.Region.EndKey, []byte{})
 	}
+	if bytes.Compare(endKey, end) > 0 {
+		endKey = end
+	}
```
kennytm, line 385: is it possible that `endKey` is the empty slice?

breeswish:
I'm not sure what's the implementation in this PR, but a kind remember: In TiKV there are two region split methods, by size and by number of keys. When size **or** number of keys exceed the threshold the region will be splited. In lightning we'd better respect both mechanism, to avoid too many keys (which will also slow down the scanning).

lightning/backend/local.go
glorv: seems impossible, because we always split regions by each startKey and endKey of the ranges. And I update the check by using the `beforeEnd` method to make it more clear

glorv:
> I'm not sure what's the implementation in this PR, but a kind reminder: In TiKV there are two region split methods, by size and by number of keys. When size **or** number of keys exceed the threshold the region will be splited. In lightning we'd better respect both mechanism, to avoid too many keys (which will also slow down the scanning).

Thanks for the reminder. I update the split regions by take both size in bytes and key-value count into account.

lightning/backend/local.go
```diff
@@ -64,6 +64,8 @@ const (
 	gRPCBackOffMaxDelay  = 3 * time.Second
 
 	LocalMemoryTableSize = 512 << 20
+
+	regionMaxKeyCount = 1_400_000
```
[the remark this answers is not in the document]
breeswish: Looks like it comes from https://github.com/tikv/tikv/blob/71ed7b10fd494ea0be69a0258a53a0282e80d20e/etc/config-template.toml#L360

glorv:
@kennytm  @3pointer  @YuJuncen  PTAL again

lightning/backend/local.go
```diff
@@ -490,7 +507,18 @@ func (local *local) WriteToTiKV(
 		zap.Int("kv_pairs", totalCount), zap.Int64("total_bytes", size),
 		zap.Int64("buf_size", bytesBuf.totalSize()))
 
-	return leaderPeerMetas, nil
+	var remainRange *Range
+	if iter.Valid() && iter.Next() {
+		firstKey = make([]byte, len(iter.Key()))
+		copy(firstKey, iter.Key())
```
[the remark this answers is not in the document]
kennytm: (still not fixed)

lightning/backend/local.go
```diff
@@ -498,7 +516,18 @@ func (local *local) WriteToTiKV(
...
+		remainRange = &Range{start: firstKey, end: endKey}
+		log.L().Info("write to tikv partial finish", zap.Int("count", totalCount),
+			zap.Int64("size", size), zap.Binary("startKey", startKey), zap.Binary("endKey", endKey),
+			zap.Reflect("remainStart", remainRange.start), zap.Reflect("remainEnd", remainRange.end),
```
3pointer, line 526: zap.Binary("remainStart", remainRange.start), zap.Binary("remainEnd", remainRange.end),

lightning/backend/local.go
```diff
@@ -977,29 +975,64 @@ func (local *local) WriteAndIngestByRanges(ctx context.Context, engineFile *Loca
...
+			return err
+		}
+		// start to write to kv and ingest
+		err = local.WriteAndIngestByRanges(ctx, engineFile.(*LocalFile), ranges, remains)
+		if err != nil {
+			log.L().Error("write and ingest ranges failed", zap.Error(err))
+			return err
+		}
+
+		unfinishedRanges := remains.take()
```
3pointer, line 1029: Maybe add a debug log here how many ranges is remain.

3pointer (commented):
Rest LGTM

kennytm (commented):
LGTM

3pointer (approved):
LGTM

PR #375, opened 2020-08-24
Fix running unit tests on Windows

kennytm:

Support running unit tests on Windows again.

1. Only enable the RLimit check for Local backend on Unix, so Lightning can be built on Windows.
2. Inside `listFiles`, always convert make path use `/` as separator

### Check List 

Tests 

 - Unit test

Side effects

Related changes


glorv (approved):
LGTM

3pointer (approved):
LGTM

PR #377, opened 2020-08-25
restore: better estimate task remain time progress log

### LINKED ISSUE #303: Progress is inaccurate (overestimating time taken) when reusing checkpoint
Repro steps:

1. Import some large database, long enough for the progress log to appear
2. Ctrl+C before all tables are written
3. Resume from checkpoint
4. Note that the remaining time is very long, as it includes those files already processed.

glorv:
Skip chunks that are finished restore when calculate estimate remain chunks numbers.  This can help make progress log a little more accurate. 

NOTE: the currently progress log do not consider the extra time needed in the import phase, so the estimate reamin time should still be lower than actual time.

maybe close #303

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Side effects

Related changes


kennytm (commented):
LGTM

lightning/restore/restore.go
```diff
@@ -403,15 +404,37 @@ func verifyCheckpoint(cfg *config.Config, taskCp *TaskCheckpoint) error {
...
 			for _, fileMeta := range tableMeta.DataFiles {
+				if cnt, ok := fileChunks[fileMeta.FileMeta.Path]; ok {
+					estimatedChunkCount += cnt
+					continue
+				}
 				if fileMeta.FileMeta.Type == mydump.SourceTypeCSV {
 					cfg := rc.cfg.Mydumper
 					if fileMeta.Size > cfg.MaxRegionSize && cfg.StrictFormat && !cfg.CSV.Header {
-						estimatedChunkCount += int(fileMeta.Size / cfg.MaxRegionSize)
+						estimatedChunkCount += math.Round(float64(fileMeta.Size) / float64(cfg.MaxRegionSize))
```
overvenus, line 437: Does it match how we split files?

glorv: yes, we split csv file by  `max-region-size` config. Each chunk maybe a little bigger than `max-region-size` since we always split chunk at the end of one line.

3pointer (approved):
LGTM

PR #379, opened 2020-08-27
backend: use pd tso as commit ts in local and importer backend

3pointer:
Before this commit we use unix timestamp as commit ts for each kv pair in tikv. 
And when we want use commits as filter(e.g. difference backup with BR), this would filter wrong result.
so I think we should encode all commit ts as pd tso.

use compose to encode commit ts.

### Check List 

Tests 

before this commit(local and importer backend are same):
屏幕快照 2020-08-27 下午5 21 52

after this commit(local and importer backend are same):
屏幕快照 2020-08-27 下午5 21 44

Related changes

### DIFF
--- a/lightning/backend/backend.go
+++ b/lightning/backend/backend.go
@@ -21,6 +21,7 @@ import (
 	"github.com/pingcap/errors"
 	"github.com/pingcap/failpoint"
 	"github.com/pingcap/parser/model"
+	"github.com/pingcap/tidb/store/tikv/oracle"
 	"github.com/pingcap/tidb/table"
 	"github.com/pingcap/tidb/types"
 	uuid "github.com/satori/go.uuid"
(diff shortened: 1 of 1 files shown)


kennytm (commented):
LGTM if test passed

glorv (approved):
LGTM

3pointer:
> LGTM if test passed

with this commit, we can backup successfully.
```
[2020/08/27 17:29:23.082 +08:00] [INFO] [collector.go:60] ["Full backup Success summary: total backup ranges: 2, total success: 2, total failed: 0, total take(s): 0.03, total kv: 2000, total size(Byte): 410000, avg speed(Byte/s): 11854717.37"] ["backup total regions"=4] [BackupTS=419042574470217732] [Size=457451]
```

and before this commit, the same command will backup nothing.

PR #382, opened 2020-09-01
test: change integration test script to allow run tests in parallel

glorv:
Currently, when we exec `TEST_NAME=... tests/run.sh`,  this will actually run all tests that name start with `TEST_NAME`. 
This pr change the meaning for `TEST_NAME` to contain one or more  test case name, and the script will only run these cases. By these way we can control the concurrency to exec the test cases. 

After this change, we can update our ci script to run integration test in parallel, see this example: https://internal.pingcap.net/idc-jenkins/blue/organizations/jenkins/lightning_ghpr_test/detail/lightning_ghpr_test/1507/pipeline/53

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Side effects

 - Change the meaning of `TEST_NAME`, so if some job depend on this, they may fail.

Related changes

### DIFF
--- a/tests/run.sh
+++ b/tests/run.sh
@@ -169,8 +169,8 @@ fi
 echo "selected test cases: $SELECTED_TEST_NAME"
 
 for casename in $SELECTED_TEST_NAME; do
-  script=tests/$casename/run.sh
-  echo "\x1b[32;1m@@@@@@@ Running test $script...\x1b[0m"
+    script=tests/$casename/run.sh
+    echo "\x1b[32;1m@@@@@@@ Running test $script...\x1b[0m"
     TEST_DIR="$TEST_DIR" \
     TEST_NAME="$casename" \
     CLUSTER_VERSION_MAJOR="${CLUSTER_VERSION_MAJOR#v}" \
--- a/tests/run.sh
+++ b/tests/run.sh
@@ -14,9 +14,8 @@
 # limitations under the License.
 
 set -eu
-
 TEST_DIR=/tmp/lightning_test_result
-SELECTED_TEST_NAME="${TEST_NAME-*}"
+SELECTED_TEST_NAME="${TEST_NAME-$(find tests -mindepth 2 -maxdepth 2 -name run.sh | cut -d/ -f2 | sort)}"
 export PATH="tests/_utils:$PATH"
 
 stop_services() {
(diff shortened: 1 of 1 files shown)


kennytm:
(just wanna note that pingcap/SRE#339 is still stuck in review)

glorv:
> (just wanna note that [pingcap/SRE#339](https://github.com/pingcap/SRE/pull/339) is still stuck in review)

I my experience as a CI maintainer, I prefer to directly maintain the test scripts in jenkins, this is convenient to update if needed. Not sure why we want to put them in a repo.

kennytm:
@glorv Yes direct update is convenient, but if you've changed anything wrong and lost the previous version then you're GG.

glorv:
> @glorv Yes direct update is convenient, but if you've changed anything wrong and lost the previous version then you're GG.

We have hourly backups for all the ci jobs config and unless our jenkins crashed, we can alway recover from previous build history easily if we just make wrong change in our script logic. 

BTW, I think it is a good way to maintain in a repo or anything similar, but seems the `SRE` repo is lack of maintain and I don't prefer to put these scripts in the project repo since they are tightly bind to jenkins and the facilities in our ci environments.

kennytm:
in any case if you want to run the tests in parallel a script is already ready to copy.

tests/run.sh
```diff
@@ -13,10 +13,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-set -eu
+if [ ! -z "$TEST_NAME" ]; then
+  SELECTED_TEST_NAME=$TEST_NAME
+else
+  SELECTED_TEST_NAME=$(find tests -name run.sh | grep -v tests/run.sh | sort | cut -d/  -f2 | tr '\r\n' ' ')
```
kennytm, line 19:
```suggestion
  SELECTED_TEST_NAME=$(find tests -mindepth 2 -maxdepth 2 -name run.sh | cut -d/ -f2)
```

no need to convert new lines into spaces.

btw the whole thing can be simply

```sh
SELECTED_TEST_NAME="${TEST_NAME-$(find tests -mindepth 2 -maxdepth 2 -name run.sh | cut -d/ -f2)}"
```

tests/run.sh
```diff
@@ -13,10 +13,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-set -eu
+if [ ! -z "$TEST_NAME" ]; then
```
kennytm, line 16:
```suggestion
if [ -n "$TEST_NAME" ]; then
```

kennytm (commented):
rest lgtm

tests/run.sh
```diff
@@ -167,12 +166,16 @@ if [ "${1-}" = '--debug' ]; then
     read line
 fi
 
-for script in tests/$SELECTED_TEST_NAME/run.sh; do
-    echo "\x1b[32;1m@@@@@@@ Running test $script...\x1b[0m"
+echo "selected test cases: $SELECTED_TEST_NAME"
+
+for casename in $SELECTED_TEST_NAME; do
+  script=tests/$casename/run.sh
+  echo "\x1b[32;1m@@@@@@@ Running test $script...\x1b[0m"
```
overvenus, line 173:
```suggestion
    script=tests/$casename/run.sh
    echo "\x1b[32;1m@@@@@@@ Running test $script...\x1b[0m"
```

overvenus (approved):
LGTM

3pointer (commented):
LGTM

3pointer (approved):

glorv:
seems after this https://github.com/pingcap/tidb/pull/19688, the cluster index is enabled by default, so we must manual disable it for the tests that depend on `_tidb_row_id`

glorv:
@kennytm  shall we merge this now?

3pointer (approved):

PR #383, opened 2020-09-02
Check Lightning version when reusing checkpoint

kennytm:

The checkpoint file is not supposed to be reused across multiple Lightning versions. Attempting to reuse checkpoint, even for stale ones, may result in non-obvious errors like https://asktug.com/t/topic/36932/18.

Also record the Lightning release version into the task checkpoint. Throw an error if the current Lightning version and the checkpoint version differ. This check can be disabled with `--check-requirements=false` as usual.

(Additionally, reduced the log level of file-router from Info to Debug since it produced too many noise.)

### Check List 

Tests 

 - Unit test

Side effects

Related changes


glorv (approved):
LGTM

overvenus (approved):
LGTM

PR #386, opened 2020-09-03
checkpoint: wait checkpoint finished if exit before success

glorv:
Fix the bug that if lightning exit before finished, e.g. manually cancelled, some checkpoint update may not be written to db before exit.

We have encounter this bug several times in recent integration test's failure. Example:  https://internal.pingcap.net/idc-jenkins/blue/organizations/jenkins/lightning_ghpr_test/detail/lightning_ghpr_test/1570/pipeline/47/

If lightning exit early, wait for checkpoint finish before exit.

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Side effects

Related changes


kennytm (commented):
LGTM

PR #387, opened 2020-09-03
backend: fix sample when split region size is small

glorv:
fix some bug for split region with sample:
- when split data size is too small, directly use naive split logic
- skip write and ingest kv if the range is empty. This can happen when a range is across multi region but kvs for some region are empty
- change script `check_cluster_verion` to support equal, because currently the cluster version for master is alway `v4.0.0` 
- add a `sleep 2` after set `@@global.tidb_enable_clustered_index = 0` to wait for cache invalidate
- change the SlowDownImport sleep time to 250ms in `checkpoint` test, since this test is too slow, the origin test will cost almost 9min to run all three backends, longer than total time of all the other tests

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Side effects

Related changes


kennytm (commented):
LGTM

lightning/backend/local.go
```diff
@@ -390,7 +390,11 @@ func (local *local) WriteToTiKV(
 	iter := engineFile.db.NewIter(opt)
 	defer iter.Close()
 
-	iter.First()
+	if !iter.First() {
+		log.L().Warn("keys within region is empty, skip ingest")
```
3pointer, line 394: how could this happen if we use paginateScanRegion to get regions?

3pointer (approved):
LGTM

kennytm (approved):

PR #396, opened 2020-09-11
checksum: use gc ttl api for checksum gc safepoint in v4.0 cluster

### LINKED ISSUE #389: Use GC-TTL to keep the GC Lifetime
## Feature Request

Currently we use `UPDATE mysql.tidb SET VARIABLE_VALUE = '100h' WHERE VARIABLE_NAME = 'tikv_gc_life_time'` to disable the GC. This is error-prone when Lightning exits before this can be reverted.

Use the more recent GC-TTL API so that PD can automatically re-enable GC even when Lightning exits accidentally.

glorv:
close #389

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Side effects

Related changes


kennytm:
IIRC @overvenus suggested that the GC-TTL API cannot be used to affect the TiDB GC lifetime. Could you confirm with the TiDB team whether this change is effective? Thanks.

overvenus:
Confirmed with @MyonKeminta, service GC safe point (GC TTL) should resolve this problem.

overvenus (commented):
Please resolve conflicts and add a unit test.

lightning/restore/restore.go
```diff
@@ -793,11 +804,47 @@ func (rc *RestoreController) restoreTables(ctx context.Context) error {
...
+func (rc *RestoreController) newChecksumManager() (ChecksumManager, error) {
+	pdAddr := rc.cfg.TiDB.PdAddr
+	pdVersion, err := common.FetchPDVersion(rc.tls, pdAddr)
+	if err != nil {
+		return nil, errors.Trace(err)
+	}
+
+	// for v4.0.0 or upper, we can use the gc ttl api
+	var manager ChecksumManager
+	if pdVersion.Major >= 4 {
```
3pointer, line 821: I remember it's not all 4.0 version support gc-ttl...

glorv: since `v4.0.0.rc.1` pd has supported this interface, so I think this check is ok?

lightning/restore/restore.go
```diff
@@ -1742,6 +1815,176 @@ func increaseGCLifeTime(ctx context.Context, db *sql.DB) (err error) {
...
+	return err
+}
+
+func (m *gcTTLManager) start(ctx context.Context) {
+	// It would be OK since TTL won't be zero, so gapTime should > `0.
+	updateGapTime := time.Duration(serviceSafePointTTL) * time.Second / preUpdateServiceSafePointFactor
+
+	updateTick := time.NewTicker(updateGapTime)
+
+	_ = m.updateGCTTL(ctx)
```
3pointer, line 1969: How about add a log for potential error?

lightning/restore/restore.go
```diff
@@ -1742,6 +1815,176 @@ func increaseGCLifeTime(ctx context.Context, db *sql.DB) (err error) {
...
+	go func() {
+		defer updateTick.Stop()
+		for {
+			select {
+			case <-ctx.Done():
+				log.L().Info("service safe point keeper exited")
+				return
+			case <-updateTick.C:
+				if err := m.updateGCTTL(ctx); err != nil {
+					log.L().Warn("failed to update service safe point, backup may fail if gc triggered",
```
3pointer, line 1979: backup?

3pointer (commented):
> LGTM, but why not use one service ts point for checksum. not per table.
lightning only set service gc safepoint we run table checksum, currently approach only set gc safepoint when start checksum. So the safepoint should be always the newest.  if lightning will run many hours, the safe point can be updated on time

3pointer (commented):
LGTM

lightning/restore/restore.go
```diff
@@ -858,11 +870,47 @@ func (rc *RestoreController) restoreTables(ctx context.Context) error {
...
 	logTask.End(zap.ErrorLevel, err)
 	return err
 }
 
+func (rc *RestoreController) newChecksumManager() (ChecksumManager, error) {
+	pdAddr := rc.cfg.TiDB.PdAddr
+	pdVersion, err := common.FetchPDVersion(rc.tls, pdAddr)
+	if err != nil {
+		return nil, errors.Trace(err)
+	}
```
kennytm, line 883: fetch PD version will fail when using "TiDB backend" outside of the cluster (i.e. the pd-addr is inaccessible).

you should return a no-op manager for TiDB backend instead of an error.

lightning/restore/restore.go
```diff
@@ -1696,23 +1744,35 @@ type RemoteChecksum struct {
...
 }
 
-// DoChecksum do checksum for tables.
-// table should be in <db>.<table>, format.  e.g. foo.bar
-func DoChecksum(ctx context.Context, db *sql.DB, table string) (*RemoteChecksum, error) {
-	var err error
-	manager, ok := ctx.Value(&gcLifeTimeKey).(*gcLifeTimeManager)
-	if !ok {
-		return nil, errors.New("No gcLifeTimeManager found in context, check context initialization")
+type ChecksumManager interface {
```
kennytm, line 1747: perhaps all two entire checksum managers into another file. this file is too large 😂

lightning/restore/restore.go
glorv: fixed, if checksum is not enabled, return nil instead

kennytm (approved):
LGTM

PR #405, opened 2020-09-21
backend: always retry ingest and get region if it's retryble

glorv:
Retry  ingest sst files and GetRegion as many times as we can to make ingest more stable

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Side effects

Related changes


glorv:
@3pointer  @YuJuncen  @kennytm  PTAL

lightning/backend/local.go
```diff
@@ -1127,8 +1134,23 @@ func isIngestRetryable(resp *sst.IngestResponse, region *split.RegionInfo, meta
 				Leader: newLeader,
 				Region: region.Region,
 			}
-			return true, newRegion, errors.Errorf("not leader: %s", errPb.GetMessage())
+		} else {
+			var err error
+			for i := 0; ; i++ {
```
kennytm, line 1139: unlimited retry? 🤔

glorv: yes, I think it's make sense to retry until GetRegion return non-nil value or an error.

This strategy is the same as br at: https://github.com/pingcap/br/blob/5af97a1a579f0f54f61c191081611bd92058d15e/pkg/restore/import.go#L260-L281

kennytm (commented):
LGTM

YuJuncen:
LGTM

kennytm (approved):

PR #411, opened 2020-09-28
config: fix error on `-d 'C:\Windows\Path'`

kennytm:

On Windows, running Lightning with `-d 'C:\Windows\Path'` would fail with "Unsupported data-source-dir url", because the config interpreted it as an URL with scheme "C".

Also the unit tests were failing on Windows.

Check if the path is an absolute Windows path before parsing the input as an URL.

Added appropriate `ToSlash` and `FromSlash` in the unit tests.

### Check List 

Tests 

 - Unit test

Side effects

Related changes


glorv (approved):
LGTM

overvenus:
LGTM

PR #414, opened 2020-10-01
mydump: fix infinite loop in ExportStatement when Read() returns an error but is not EOF

kennytm:

In the `ExportStatement()` function, if the reader errored which is not EOF, it will cause the function go into an infinite loop because the error is ignored.

Handle the error.

### Check List 

Tests 

 - Unit test

Side effects

Related changes


july2993 (approved):

glorv:
LGTM

PR #415, opened 2020-10-01
lightning: start the HTTP server when receiving SIGUSR1

kennytm:

Allow user to dynamically start the HTTP server even if it was disabled, to allow emergency use of PProf.

Added a SIGUSR1 listener, which will start the server listening from a random port when signaled (like pingcap/br#372).

### Check List 

Tests 

    * `kill -USR1 pid` and check.

Side effects

Related changes


glorv (approved):
LGTM

3pointer (commented):
LGTM

PR #417, opened 2020-10-09
support restore view

### LINKED ISSUE #297: Support restoring VIEWs and SEQUENCEs
## Feature Request

TiDB and MariaDB both support VIEW and SEQUENCE, but Lightning is unable to import them yet. This makes restoring/migrating an SQL backup tedious, as the VIEW and SEQUENCE have to be manually reconstructed.

Support restoring VIEW and SEQUENCE, or its general form, pre-create-table SQLs and post-create-table SQLs.

SEQUENCEs must be created before tables are created, because tables may use `NEXT VALUE FOR seq` as the default value.

VIEWs must be created after tables are created, since the structure of the views depend on the tables (in mydumper and mysqldump, VIEWs are first created as tables to define the name and structure, and then dropped and replaced by the real VIEW definition.)

-

-

glorv:
support restore view. parcially resolve #297 

- file routing add a new file type `view-schema` for view, default routing pattern is `{schema}.{table}-schema-view.sql`
- restore view in `restoreSchema` after all table schemas are restored

NOTE: Currently, we only support the view schema files exported by [dumpling](https://github.com/pingcap/dumpling) like tools. That is, for a table `db.tbl`, there must exist two schema files:
- db.tbl-schema.sql. This is for the stub table schema file.
- db.tbl-schema-view.sql. This is the actual view schema file.

If the stub table schema file does not exist, currently, lightning will report an error and exit directly.

WARN: This pr only handles the name mapping for the view name but not for the table name in the queries in the `CREATE VIEW` statement. 

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Side effects

Related changes


tests/view/run.sh
```diff
@@ -0,0 +1,82 @@
...
+SET @PREV_CHARACTER_SET_RESULTS=@@CHARACTER_SET_RESULTS;
+SET @PREV_COLLATION_CONNECTION=@@COLLATION_CONNECTION;
+SET character_set_client = utf8;
+SET character_set_results = utf8;
+SET collation_connection = utf8_general_ci;
+CREATE ALGORITHM=UNDEFINED DEFINER=`root`@`192.168.198.178` SQL SECURITY DEFINER VIEW `v2` (`s`) AS SELECT `s` FROM `db1`.`v1` WHERE `i`<2;
+SET character_set_client = @PREV_CHARACTER_SET_CLIENT;
+SET character_set_results = @PREV_CHARACTER_SET_RESULTS;
+SET collation_connection = @PREV_COLLATION_CONNECTION;
+_EOF_
```
kennytm, line 61: why not use static files for all of these

lightning/restore/restore.go
```diff
@@ -1166,8 +1184,9 @@ func (t *TableRestore) importEngine(
 }
 
 func (t *TableRestore) postProcess(ctx context.Context, rc *RestoreController, cp *TableCheckpoint) error {
-	if !rc.backend.ShouldPostProcess() {
-		t.logger.Debug("skip post-processing, not supported by backend")
+	// if post-process is disabled or the table is empty, just skip
+	if !rc.backend.ShouldPostProcess() || len(cp.Engines) == 1 {
```
kennytm, line 1188: should it be `len() > 0` here

kennytm (commented):
Rest LGTM

lightning/restore/restore.go
glorv: `len(cp.Engines) == 1` means there is only one opened engine -- the index engine, thus there are no kv in this engine, so it's safe to just skip post process.

kennytm: okay.

lightning/restore/tidb.go
```diff
@@ -137,52 +137,69 @@ func InitSchema(ctx context.Context, g glue.Glue, database string, tablesSchema
...
 
-		err = sqlExecutor.ExecuteWithLog(
-			ctx,
-			sqlCreateTable,
-			"create table",
-			logger.With(zap.String("table", common.UniqueTable(database, tbl))),
-		)
-		if err != nil {
-			break
+		//TODO: maybe we should put these createStems into a transaction
```
lance6716, line 149: not sure we could do this. seems DDL will do "an implicit commit after executing"

https://dev.mysql.com/doc/refman/5.6/en/implicit-commit.html

lightning/restore/tidb.go
```diff
@@ -137,52 +137,69 @@ func InitSchema(ctx context.Context, g glue.Glue, database string, tablesSchema
...
 	}
 
 	var res strings.Builder
-	res.Grow(len(createTable))
 	ctx := format.NewRestoreCtx(format.DefaultRestoreFlags, &res)
 
+	retStmts := make([]string, 0, len(stmts))
 	for _, stmt := range stmts {
-		if createTableNode, ok := stmt.(*ast.CreateTableStmt); ok {
+		switch stmt.(type) {
```
lance6716, line 178:
```suggestion
		switch n := stmt.(type) {
```

lance6716 (commented):
rest LGTM

lichunzhu (approved):
LGTM

lightning/restore/tidb.go
glorv: These actions are not all DDLs. Some of them are session variables setting.

lance6716 (approved):

PR #419, opened 2020-10-12
backend/local: do not retry epochNotMatch error when ingest sst

glorv:
- add retry for `SplitAndScatterRegionByRanges` since it may caused by region leader changes
- do not directly retry ingest if the return error is `EpochNotMatch` because tikv will check the epoch in SstMeta and the newest Region info, but we can change the epoch in SstMeta 
- retry ingest if the error is `Raft raft: proposal dropped`, because this means there is a leader transfer for this region
- downgrade some log level from Error to Warn, since they are retryable
- fix a bug in backend/log that cause scan region by range return one more region that does not intersect with the range  if the range end key is prefix of the region start key

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Side effects

Related changes

Release Note
- fix several trivial bugs when ingesting sst files in local backend

### DIFF
--- a/lightning/backend/local.go
+++ b/lightning/backend/local.go
@@ -1175,6 +1175,8 @@ func (local *local) isIngestRetryable(
 	return false, nil, errors.Errorf("non-retryable error: %s", resp.GetError().GetMessage())
 }
 
+// return the smallest []byte that is bigger than current bytes.
+// special case when key is empty, empty bytes means infinity in our context, so directly return itself.
 func nextKey(key []byte) []byte {
 	if len(key) == 0 {
 		return []byte{}
--- a/lightning/backend/local.go
+++ b/lightning/backend/local.go
@@ -1240,7 +1240,7 @@ func (local *local) isIngestRetryable(
 			}
 		}
 		return true, newRegion, errors.Errorf("not leader: %s", errPb.GetMessage())
-	case strings.Contains(errPb.Message, "Raft raft: proposal dropped"):
+	case strings.Contains(errPb.Message, "raft: proposal dropped"):
 		// TODO: we should change 'Raft raft: proposal dropped' to a error type like 'NotLeader'
 		newRegion, err = getRegion()
 		if err != nil {
--- a/lightning/backend/local.go
+++ b/lightning/backend/local.go
@@ -1009,11 +1009,17 @@ func (local *local) ImportEngine(ctx context.Context, engineUUID uuid.UUID) erro
 
 	for {
 		// split region by given ranges
-		err = local.SplitAndScatterRegionByRanges(ctx, ranges)
+		for i := 0; i < maxRetryTimes; i++ {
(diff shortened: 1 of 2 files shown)


3pointer (commented):
LGTM

lightning/backend/local.go
```diff
@@ -1211,65 +1234,32 @@ func (local *local) isIngestRetryable(
...
-	for i := len(key) - 1; i >= 0; i-- {
-		if key[i] != '\xff' {
-			pos = i
-			break
-		}
-	}
-	s, e := key[:pos], key[pos]+1
-	res = append(append(res, s...), e)
+	res = append(res, key...)
+	res = append(res, 0)
```
overvenus, line 1262: Could you add a unit test?

lightning/backend/local.go
```diff
@@ -1211,65 +1234,32 @@ func (local *local) isIngestRetryable(
...
-					}
-				}
-				if newLeader != nil {
-					newRegion = &split.RegionInfo{
-						Leader: newLeader,
-						Region: currentRegion,
-					}
-				}
-			}
+	case strings.Contains(errPb.Message, "Raft raft: proposal dropped"):
```
overvenus, line 1243:
```suggestion
	case strings.Contains(errPb.Message, "raft: proposal dropped"):
```

Also please add some comments.

overvenus (commented):
Rest LGTM

overvenus:
Please add release note.

lightning/backend/local.go
```diff
@@ -1202,7 +1209,23 @@ func (local *local) isIngestRetryable(
...
+			newRegion, err := local.splitCli.GetRegion(ctx, region.Region.GetStartKey())
+			if err != nil {
+				return nil, errors.Trace(err)
+			}
+			if newRegion != nil {
+				return newRegion, nil
+			}
+			log.L().Warn("get region by key return nil, will retry", zap.Reflect("region", region),
+				zap.Int("retry", i))
+			time.Sleep(time.Second)
```
kennytm, line 1223: select this `time.Sleep` with a `<-ctx.Done()` to avoid Ctrl+C being unresponsive

kennytm (commented):
rest LGTM

lightning/backend/local.go
```diff
@@ -1220,7 +1220,12 @@ func (local *local) isIngestRetryable(
 			}
 			log.L().Warn("get region by key return nil, will retry", zap.Reflect("region", region),
 				zap.Int("retry", i))
-			time.Sleep(time.Second)
+			select {
+			case <-ctx.Done():
+				return nil, ctx.Err()
+			default:
+				time.Sleep(time.Second)
```
kennytm, line 1227:
```suggestion
			case <-time.After(time.Second):
```

overvenus (approved):
LGTM

PR #425, opened 2020-10-22
log: simplify some warn log and do retry write for epoch not match error

glorv:
- Simplify some error and warn log. Downgrade the log level or change zap.Error to log.ShortError to make log more user friendly
- When ingesting returns epoch-not-match error, if the range still in the new region, we should retry from writing kv-pairs instead of directly return an error. The updated logic should be the same as current tikv-import implementation

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Side effects

Related changes

### DIFF
--- a/lightning/backend/local.go
+++ b/lightning/backend/local.go
@@ -1022,7 +1022,7 @@ loopWrite:
 		return remainRange, errors.Trace(err)
 	}
 
-	return remainRange, err
+	return remainRange, errors.Trace(err)
 }
 
 func (local *local) WriteAndIngestByRanges(ctx context.Context, engineFile *LocalFile, ranges []Range, remainRanges *syncdRanges) error {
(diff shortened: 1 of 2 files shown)


kennytm:
LGTM

lightning/backend/local.go
```diff
@@ -941,69 +945,83 @@ WriteAndIngest:
...
+						resp.Error.EpochNotMatch = &errorpb.EpochNotMatch{
+							CurrentRegions: []*metapb.Region{region.Region}}
+					}
+				})
+				var retryTy retryType
+				var newRegion *split.RegionInfo
+				retryTy, newRegion, err = local.isIngestRetryable(ctx, resp, region, meta)
+				if err == nil {
+					// ingest next meta
+					break
```
3pointer, line 1000: when break `loopWrite`?

glorv: fixed

lightning/backend/local.go
```diff
@@ -941,70 +945,84 @@ WriteAndIngest:
...
-			log.L().Warn("all retry ingest failed", zap.Reflect("ingest meta", meta), zap.Error(err))
-			return remainRange, errors.Trace(err)
+			log.L().Warn("write and ingest region, will retry import full range", log.ShortError(err),
+				zap.Stringer("region", region.Region), zap.Binary("start", start), zap.Binary("end", end))
 		}
+		return remainRange, errors.Trace(err)
 	}
-	return remainRange, nil
+
+	return remainRange, err
```
3pointer, line 1025:
```suggestion
	return remainRange, errors.Trace(err)
```

3pointer (commented):
Rest LGTM

3pointer (approved):
LGTM

PR #430, opened 2020-10-27
backend/tidb: fix compatible issue in fetch table schemas 

glorv:
Fix the compatible issue in `show table t next_row_id` with v4.0.0~ v4.0.2

In v4.0.0 tidb  `show table t next_row_id` don't return the `ID_TYPE ` column, so we should Scan by column count and set the default value without the extra row

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Side effects

Related changes


overvenus:
LGTM

kennytm (approved):
LGTM

PR #441, opened 2020-11-03
config: support encode PostOpLevel and Duration as input

lance6716:
currently in lightning via SQL, task config is encoded to string and write to `GlobalConfig.ConfigFileContent`. (I'll add a better interface later). `PostOpLevel` and `Duration` are different from their input format after encoding.

add `MarshalText` to these struct

### Check List 

Tests 

 - Unit test

Side effects

Related changes

### DIFF
--- a/lightning/config/config.go
+++ b/lightning/config/config.go
@@ -145,6 +145,12 @@ func (t *PostOpLevel) UnmarshalTOML(v interface{}) error {
 		}
 	case string:
 		return t.FromStringValue(val)
+	case int64:
+		if int64(OpLevelOff) <= val && val <= int64(OpLevelRequired) {
+			*t = PostOpLevel(val)
+		} else {
+			return errors.Errorf("invalid op level '%v', please choose valid option between ['off', 'optional', 'required']", v)
+		}
 	default:
 		return errors.Errorf("invalid op level '%v', please choose valid option between ['off', 'optional', 'required']", v)
 	}
--- a/lightning/config/config.go
+++ b/lightning/config/config.go
@@ -145,16 +145,14 @@ func (t *PostOpLevel) UnmarshalTOML(v interface{}) error {
 		}
 	case string:
 		return t.FromStringValue(val)
-	case int64:
-		if int64(OpLevelOff) <= val && val <= int64(OpLevelRequired) {
-			*t = PostOpLevel(val)
-		} else {
-			return errors.Errorf("invalid op level '%v', please choose valid option between ['off', 'optional', 'required']", v)
-		}
 	default:
 		return errors.Errorf("invalid op level '%v', please choose valid option between ['off', 'optional', 'required']", v)
 	}
 	return nil
+}
+
+func (t PostOpLevel) MarshalText() ([]byte, error) {
+	return []byte(t.String()), nil
 }
 
 // parser command line parameter
@@ -296,6 +294,10 @@ func (d *Duration) UnmarshalText(text []byte) error {
 	var err error
 	d.Duration, err = time.ParseDuration(string(text))
 	return err
+}
+
+func (d Duration) MarshalText() ([]byte, error) {
+	return []byte(d.String()), nil
 }
 
 func (d *Duration) MarshalJSON() ([]byte, error) {
--- a/lightning/config/config_test.go
+++ b/lightning/config/config_test.go
@@ -572,12 +572,22 @@ func (s *configTestSuite) TestTomlPostRestore(c *C) {
 		err := cfg.LoadFromTOML([]byte(confStr))
 		c.Assert(err, IsNil)
 		c.Assert(cfg.PostRestore.Checksum, Equals, v)
+
+		confStr = fmt.Sprintf("[post-restore]\r\nchecksum= %v\r\n", int(v))
+		err = cfg.LoadFromTOML([]byte(confStr))
+		c.Assert(err, IsNil)
+		c.Assert(cfg.PostRestore.Checksum, Equals, v)
 	}
 
 	for k, v := range kvMap {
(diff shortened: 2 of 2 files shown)


lightning/config/config.go
```diff
@@ -145,6 +145,12 @@ func (t *PostOpLevel) UnmarshalTOML(v interface{}) error {
 		}
 	case string:
 		return t.FromStringValue(val)
+	case int64:
```
glorv, line 148: Maybe we should implement `MarshalText` directly

lance6716:
very strange that `MarshalText` couldn't be called in lightning

```
func (d *Duration) MarshalText() ([]byte, error) {
	return []byte("11111"+d.String()), nil
}
```

```
        bb, _ := cfg.Cron.LogProgress.MarshalText()
	fmt.Println(string(bb)) // 111110s

	var buf bytes.Buffer
	err = toml.NewEncoder(&buf).Encode(cfg.Cron)
	c.Assert(err, IsNil)
	fmt.Println(buf.String()) // [log-progress]
                                  //    Duration = "0s"
                                  // [switch-mode]
                                  //   Duration = "0s"
	buf.Reset()
	err = toml.NewEncoder(&buf).Encode(cfg.Cron.LogProgress)
	c.Assert(err, IsNil)
	fmt.Println(buf.String()) // Duration = "0s"
```

kennytm:
i think we should change the parser syntax as

```sql
-- the original boolean
IMPORT ... CHECKSUM = 0;
IMPORT ... CHECKSUM = 1;
IMPORT ... CHECKSUM = FALSE;
IMPORT ... CHECKSUM = TRUE;

-- the new PostOpLevel
IMPORT ... CHECKSUM = REQUIRED;
IMPORT ... CHECKSUM = OPTIONAL;
```

users should never need to understand the difference between `IMPORT ... CHECKSUM = 2` and `IMPORT ... CHECKSUM = 1`, that's an implementation detail.

lance6716:
> i think we should change the parser syntax as
> 
> ```sql
> -- the original boolean
> IMPORT ... CHECKSUM = 0;
> IMPORT ... CHECKSUM = 1;
> IMPORT ... CHECKSUM = FALSE;
> IMPORT ... CHECKSUM = TRUE;
> 
> -- the new PostOpLevel
> IMPORT ... CHECKSUM = REQUIRED;
> IMPORT ... CHECKSUM = OPTIONAL;
> ```
> 
> users should never need to understand the difference between `IMPORT ... CHECKSUM = 2` and `IMPORT ... CHECKSUM = 1`, that's an implementation detail.

currently user could use `FALSE / TRUE` to represent `OFF / REQUIRED`, will update parser later.

and this PR is to solve that, TiDB construct a task config, marshall it to bytes and set to global.ConfigFileContent. This problematic conversion is between marshall and unmarshall.

BTW, lightning's `MarshalText` of
```
type Duration struct {
	time.Duration
}
```
can't be called in `toml.NewEncoder().Encode(cfg.Cron)`, any idea 😭

kennytm:
you should create the task config struct directly, and leave ConfigFileContent empty :thinking:

lance6716:
> you should create the task config struct directly, and leave ConfigFileContent empty 🤔

configurations like log-level are in global config, so if I want to change them I'll pass two config to lightning, not seems so good 🤔

kennytm:
the log level should be set as the same as tidb's (perhaps reuse tidb's logger)

lance6716:
updated, PTAL

glorv:
LGTM

lightning/config/config_test.go
```diff
@@ -566,12 +570,19 @@ func (s *configTestSuite) TestTomlPostRestore(c *C) {
...
 	for k, v := range kvMap {
 		cfg := &config.Config{}
 		confStr := fmt.Sprintf("[post-restore]\r\nchecksum= %s\r\n", k)
 		err := cfg.LoadFromTOML([]byte(confStr))
 		c.Assert(err, IsNil)
 		c.Assert(cfg.PostRestore.Checksum, Equals, v)
+
+		b.Reset()
+		c.Assert(enc.Encode(cfg.PostRestore), IsNil)
+		c.Assert(strings.Contains(b.String(), fmt.Sprintf(`checksum = "%s"`, v.String())), IsTrue)
```
kennytm, line 585:
```suggestion
		c.Assert(b, Matches, fmt.Sprintf(`.*checksum = "\Q%s\E".*`, v))
```

lightning/config/config_test.go
```diff
@@ -580,5 +591,25 @@ func (s *configTestSuite) TestTomlPostRestore(c *C) {
 		err := cfg.LoadFromTOML([]byte(confStr))
 		c.Assert(err, IsNil)
 		c.Assert(cfg.PostRestore.Analyze, Equals, v)
+
+		b.Reset()
+		c.Assert(enc.Encode(cfg.PostRestore), IsNil)
+		c.Assert(strings.Contains(b.String(), fmt.Sprintf(`analyze = "%s"`, v.String())), IsTrue)
```
kennytm, line 597: ditto

kennytm (commented):
rest LGTM

lightning/config/config_test.go
lance6716: done in https://github.com/pingcap/tidb-lightning/pull/441/commits/2f540f8c8dfafda17bb809a4f897f8b39b9a7c52

lance6716:
/run-all-tests

Cannot contact test-go1130-memvolume-624ps: hudson.remoting.RequestAbortedException: java.nio.channels.ClosedChannelException

lance6716:
CI passed, could merge now @kennytm

glorv (approved):

PR #453, opened 2020-11-05
tidb-lightning-ctl: change default of -d to 'noop://'

kennytm:

Since #361 there is a check if the data source actually exists. This breaks `tidb-lightning-ctl` which does not require the data source directory.

Change the default `-d` in `tidb-lightning-ctl` from empty string to `noop://`. 

Also added `noop` to the supported protocol list.

### Check List 

Tests 

    * --fetch-mode works without passing `-d`.

Side effects

Related changes


glorv:
LGTM

overvenus (approved):
LGTM

PR #456, opened 2020-11-06
restore: add `glue.Glue` interface and other function

lance6716:

To embed lightning into TiDB, a glue interface is needed to use host TiDB's connection. This PR moves some external TiDB usage into a glue interface, to support hiding all call to external TiDB behind glue interface with following PRs.

- add `glue.Glue` and `glue.SQLExecutor`. `glue.SQLExecutor` could do some SQL jobs.
- add some call to `Glue.Record` to report information to TiDB
- change `RunOnce` to enable using lightning as a library more friendly

### Check List 

Tests 

 - Pass original test

Side effects

 - Increased code complexity

Related changes


lightning/common/util.go
```diff
@@ -102,11 +102,11 @@ func IsEmptyDir(name string) bool {
 // SQLWithRetry constructs a retryable transaction.
 type SQLWithRetry struct {
 	DB           *sql.DB
-	Logger       log.Logger
+	Logger       *zap.Logger
```
kennytm, line 105: we could keep using `log.Logger` here. external packages just need to write `log.Logger{Logger: logger}` to wrap a `*zap.Logger`.

lance6716: this change is to enable TiDB's implementation of glue get rid of lightning's logger, not sure if there's a good way

lance6716:
/run-all-tests

last failed CI: https://internal.pingcap.net/idc-jenkins/blue/organizations/jenkins/lightning_ghpr_test/detail/lightning_ghpr_test/2521/pipeline/55

``
[2020-11-08T12:17:16.222Z] [2020/11/08 20:17:15.804 +08:00] [INFO] [lightning.go:212] [cfg] [cfg="{\"id\":1604837835795139627,\"lightning\":{\"table-concurrency\":6,\"index-concurrency\":2,\"region-concurrency\":40,\"io-concurrency\":5,\"check-requirements\":true},\"tidb\":{\"host\":\"test.invalid\",\"port\":4000,\"user\":\"root\",\"status-port\":10080,\"pd-addr\":\"test.invalid:2379\",\"sql-mode\":\"ONLY_FULL_GROUP_BY,NO_AUTO_CREATE_USER\",\"tls\":\"false\",\"security\":{\"ca-path\":\"\",\"cert-path\":\"\",\"key-path\":\"\"},\"max-allowed-packet\":67108864,\"distsql-scan-concurrency\":100,\"build-stats-concurrency\":20,\"index-serial-scan-concurrency\":20,\"checksum-table-concurrency\":16},\"checkpoint\":{\"enable\":true,\"schema\":\"tidb_lightning_checkpoint\",\"driver\":\"file\",\"keep-after-success\":false},\"mydumper\":{\"read-block-size\":65536,\"batch-size\":107374182400,\"batch-import-ratio\":0,\"data-source-dir\":\"file://demo-path-3\",\"no-schema\":false,\"character-set\":\"auto\",\"csv\":{\"separator\":\",\",\"delimiter\":\"\\\"\",\"header\":true,\"trim-last-separator\":false,\"not-null\":false,\"null\":\"\\\\N\",\"backslash-escape\":true},\"case-sensitive\":false,\"strict-format\":false,\"max-region-size\":268435456,\"filter\":[\"*.*\"],\"files\":null,\"default-file-rules\":true},\"tikv-importer\":{\"addr\":\"\",\"backend\":\"importer\",\"on-duplicate\":\"replace\",\"max-kv-pairs\":32,\"send-kv-pairs\":32768,\"region-split-size\":100663296,\"sorted-kv-dir\":\"\",\"range-concurrency\":16},\"post-restore\":{\"level-1-compact\":false,\"compact\":false,\"checksum\":\"required\",\"analyze\":\"optional\"},\"cron\":{\"switch-mode\":\"5m0s\",\"log-progress\":\"5m0s\"},\"routes\":null,\"security\":{\"ca-path\":\"\",\"cert-path\":\"\",\"key-path\":\"\"},\"black-white-list\":{\"do-tables\":null,\"do-dbs\":null,\"ignore-tables\":null,\"ignore-dbs\":null}}"]
[2020-11-08T12:17:16.223Z] 
[2020-11-08T12:17:16.223Z] ----------------------------------------------------------------------
[2020-11-08T12:17:16.223Z] FAIL: lightning_test.go:191: lightningServerSuite.TestGetDeleteTask
[2020-11-08T12:17:16.223Z] 
[2020-11-08T12:17:16.223Z] lightning_test.go:336:
[2020-11-08T12:17:16.223Z]     c.Assert(getAllTasks(), DeepEquals, getAllResultType{
[2020-11-08T12:17:16.223Z]         Current: third,
[2020-11-08T12:17:16.223Z]         Queue:   []int64{},
[2020-11-08T12:17:16.223Z]     })
[2020-11-08T12:17:16.223Z] ... obtained lightning.getAllResultType = lightning.getAllResultType{Current:0, Queue:[]int64{}}
[2020-11-08T12:17:16.223Z] ... expected lightning.getAllResultType = lightning.getAllResultType{Current:1604837835795139627, Queue:[]int64{}}
[2020-11-08T12:17:16.223Z]
```

**Comment by lance6716** [comment]
/run-all-tests

my local s3 test with minio version RELEASE.2020-11-06T23-17-07Z could pass 🤔

File: lightning/glue/glue.go [hunk 2]
```diff
@@ -0,0 +1,92 @@
...
+
+type SQLExecutor interface {
+	ExecuteWithLog(ctx context.Context, query string, purpose string, logger *zap.Logger) error
+	ObtainStringLog(ctx context.Context, query string, purpose string, logger *zap.Logger) (string, error)
+	Close()
+}
+
+type ExternalTiDBGlue struct {
+	db 	   *sql.DB
+	parser *parser.Parser
```
kennytm, line 43: fix format

lightning/restore/tidb.go
```diff
@@ -302,14 +304,11 @@ func AlterAutoIncrement(ctx context.Context, db *sql.DB, tableName string, incr
 	return errors.Annotatef(err, "%s", query)
 }
 
-func AlterAutoRandom(ctx context.Context, db *sql.DB, tableName string, randomBase int64) error {
-	sql := common.SQLWithRetry{
-		DB:     db,
-		Logger: log.With(zap.String("table", tableName), zap.Int64("auto_random", randomBase)),
-	}
+func AlterAutoRandom(ctx context.Context, g glue.SQLExecutor, tableName string, randomBase int64) error {
+	logger := log.With(zap.String("table", tableName), zap.Int64("auto_random", randomBase))
```
kennytm, line 308: will using `log.With` use TiDB's own logger?

lance6716: currently not. in following PR will add a `RunEmbeddedOnce(context.Context, config.Config, zap.Logger)` that replace `appLogger` with given logger

lance6716:
going to let TiDB implement a "glue" `CheckpointsDB`, here's the change of glue interface follwing this PR.
https://github.com/lance6716/tidb-lightning/pull/1/files

PTAL @overvenus 

and PTAL for this PR @kennytm @glorv

tests/s3/run.sh
```diff
@@ -28,7 +28,7 @@ DBPATH="$TEST_DIR/s3.mydump"
 export MINIO_ACCESS_KEY=s3accesskey
 export MINIO_SECRET_KEY=s3secretkey
 export MINIO_BROWSER=off
-export S3_ENDPOINT=127.0.0.1:9000
+export S3_ENDPOINT=127.0.0.1:9900
```
overvenus, line 31: Why change port?

lightning/glue/glue.go
```diff
@@ -0,0 +1,120 @@
...
+	GetTables(context.Context, string) ([]*model.TableInfo, error)
+	OpenCheckpointsDB(context.Context, *config.Config) (checkpoints.CheckpointsDB, error)
+	Record(string, uint64)
+}
+
+type SQLExecutor interface {
+	ExecuteWithLog(ctx context.Context, query string, purpose string, logger *zap.Logger) error
+	ObtainStringLog(ctx context.Context, query string, purpose string, logger *zap.Logger) (string, error)
+	Close()
+}
```
overvenus, line 43: Why pass a logger through the interface?

What about passing a logger when we creating an SQLExecutor instance?

```golang
func NewExternalTiDBGlue(db *sql.DB, sqlMode mysql.SQLMode, logger *zap.Logger) *ExternalTiDBGlue { ... }
```

tests/s3/run.sh
lance6716: some CI keep failing because of port conflict (see this PR's history)

lightning/glue/glue.go
lance6716: some functions need a logger from `Task`, and that `Task` is often built from logger with different fields. It seems that pinning a logger at `New` is not very flexible

lance6716:
following glue checkpoint change: https://github.com/lance6716/tidb-lightning/pull/2/files

PTAL for this PR @overvenus @kennytm @glorv @3pointer

lightning/lightning.go
```diff
@@ -185,6 +186,19 @@ func (l *Lightning) RunOnce() error {
 	return l.run(cfg)
 }
 
+func (l *Lightning) RunEmbeddedOnce(ctx context.Context, taskCfg *config.Config, logger *zap.Logger, g glue.Glue) error {
```
glorv, line 189: This method doesn't have much difference with `RunOnce`, we reuse the `RunOnce` directly with little change?

lightning/lightning.go
```diff
@@ -207,7 +221,7 @@ func (l *Lightning) RunServer() error {
 
 var taskCfgRecorderKey struct{}
 
-func (l *Lightning) run(taskCfg *config.Config) (err error) {
+func (l *Lightning) run(taskCfg *config.Config, glues ...glue.Glue) (err error) {
```
glorv, line 224: Are there any use cases that need multi glues?

lightning/log/log.go
```diff
@@ -106,6 +106,11 @@ func L() Logger {
 	return appLogger
 }
 
+// SetAppLogger replaces the default logger in this package to given one
+func SetAppLogger(l *zap.Logger) {
+	appLogger = Logger{l.WithOptions(zap.AddStacktrace(zap.DPanicLevel))}
```
glorv, line 111: Why still need to change the log config here?

lightning/restore/checksum.go
```diff
@@ -84,7 +85,11 @@ func newChecksumManager(rc *RestoreController) (ChecksumManager, error) {
 
 		manager = newTiKVChecksumManager(store.(tikv.Storage).GetClient(), pdCli)
 	} else {
-		manager = newTiDBChecksumExecutor(rc.tidbMgr.db)
+		e, ok := rc.tidbGlue.GetSQLExecutor().(glue.ExternalTiDBGlue)
```
glorv, line 88: should TiDBChecksumExecutor use Glue directly here directly? Why we need to care about the implementation of glue here?

lightning/restore/restore.go
```diff
@@ -238,35 +241,11 @@ func NewRestoreControllerWithPauser(
...
-
-	default:
-		return nil, errors.Errorf("Unknown checkpoint driver %s", cfg.Checkpoint.Driver)
-	}
-}
-
 func (rc *RestoreController) Close() {
 	rc.backend.Close()
-	rc.tidbMgr.Close()
+	if rc.tidbGlue.OwnsSQLExecutor() {
```
glorv, line 246: why not implement a `Glue.Close` to avoid doing the `own` check?

lightning/restore/restore.go
```diff
@@ -333,15 +314,19 @@ func (rc *RestoreController) restoreSchema(ctx context.Context) error {
...
 
 			task.End(zap.ErrorLevel, err)
 			if err != nil {
 				return errors.Annotatef(err, "restore table schema %s failed", dbMeta.Name)
 			}
 		}
 	}
-	dbInfos, err := tidbMgr.LoadSchemaInfo(ctx, rc.dbMetas, rc.backend.FetchRemoteTableModels)
+	getTableFunc := rc.backend.FetchRemoteTableModels
+	if !rc.tidbGlue.OwnsSQLExecutor() {
```
glorv, line 326: I think glue should provide `FetchRemoteTableModels` here

lightning/lightning.go
lance6716: > Are there any use cases that need multi glues?

No, just to make old `run(cfg)` could work

lightning/log/log.go
lance6716: keep same with https://github.com/pingcap/tidb-lightning/blob/963829fd54f16ba135790bc8bde4b8c2d48fb1aa/lightning/log/log.go#L96-L98

lightning/restore/checksum.go
lance6716: as below error message, if a TiDB (with lightning) runs with PD which version is less than v4.0.0, this function will call `newTiDBChecksumExecutor` which need a `*sql.DB` that only `ExternalTiDBGlue` owns. I want to give a friendly message about this scenario

lightning/restore/restore.go
lance6716: at where lightning's glue is initiated, backend is not initiated. So I guess we could

- keep this implementation, because `OwnsSQLExecutor` has another usage that decide whether run `SET SQL_MODE =`
- decide backend then initiate glue
- initiate backend afterwards, and `glue.SetFetchRemoteTableModels`

what do you think

lightning/lightning.go
kennytm: `run` is a private function you could change all invocation to do `.run(cfg, nil)` or something

lightning/lightning.go
lance6716: https://github.com/pingcap/tidb-lightning/pull/456/commits/6e4da0a4257a061a0882b3de76b3cafb661f0682

lightning/lightning.go
lance6716: https://github.com/pingcap/tidb-lightning/pull/456/commits/6e4da0a4257a061a0882b3de76b3cafb661f0682

lightning/restore/restore.go
lance6716: https://github.com/pingcap/tidb-lightning/pull/456/commits/832f65cf8877c1b9228721f86bb77e177a786de8

lightning/lightning.go
```diff
@@ -229,17 +235,39 @@ func (l *Lightning) run(taskCfg *config.Config) (err error) {
...
 			case <-ctx.Done():
-				return ctx.Err()
+				failpoint.Return(ctx.Err())
 			}
 		}
-		return nil
+		failpoint.Return(nil)
 	})
 
+	if err := taskCfg.TiDB.Security.RegisterMySQL(); err != nil {
```
glorv, line 249: Are there any problems if we run multi lightning tasks with in the same tidb? Or tidb will refuse to run more than one lightning tasks?

lightning/lightning.go
```diff
@@ -284,6 +312,11 @@ func (l *Lightning) run(taskCfg *config.Config) (err error) {
 }
 
 func (l *Lightning) Stop() {
+	l.cancelLock.Lock()
```
glorv, line 315: This explicit cancel is not needed? If lightning run in server mod, the cancel task action will call this cancel? If lightning run in once mod, `l.shutdown` will stop lightning

lightning/log/log.go
glorv: They are not the same.  #96~98 is the default operation to build an app log. But if we build a logger and set it as global, the caller should take responsibility to set all options and lightning shouldn't override its config anymore.

cmd/tidb-lightning/main.go
```diff
@@ -74,11 +75,18 @@ func main() {
...
-	}
+	err = func() error {
+		if globalCfg.App.ServerMode {
+			return app.RunServer()
+		} else {
+			cfg := config.NewConfig()
+			if err := cfg.LoadFromGlobal(globalCfg); err != nil {
+				return err
+			}
+			return app.RunOnce(context.Background(), cfg, nil, nil)
```
glorv, line 86: I have thought about it again. Since there is already a context in the `Lightning` struct `RunOnce` method can always depend on the `app.ctx` instead of another ctx as parameter?

lightning/log/log.go
lance6716: that's not override, just create a new logger for lightning's specification, and the comment remains reasonable IMO

cmd/tidb-lightning/main.go
```diff
@@ -74,11 +75,18 @@ func main() {
 		return
 	}
 
-	if cfg.App.ServerMode {
-		err = app.RunServer()
-	} else {
-		err = app.RunOnce()
-	}
+	err = func() error {
```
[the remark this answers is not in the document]
lance6716: It could be 

```golang
                        cfg := config.NewConfig()
			err = cfg.LoadFromGlobal(globalCfg)
			if err == nil {
				err = app.RunOnce(context.Background(), cfg, nil, nil)
			}
```

But I think that's more mess since `err` is assinged many place

cmd/tidb-lightning/main.go
lance6716: In future usage, lightning may be `New` once and that instance `RunOnce` for many times, like a HTTP server mode. So a task context will help

lightning/lightning.go
lance6716: currently TiDB will only run one task, and `RunOnce` is blocked thus it's unnatural to concurrently call `RunOnce` on one lightning (I hope). 

For multiple lightning in one process, `RegisterMySQL` may fail because `tlsConfigRegistry` is a global variable, I'll be careful for that

lightning/lightning.go
lance6716: A test expects this behaviour, that when lightning shutdown task should be canceled to avoid futhur actions (it's remove checkpoint in fact)

https://github.com/pingcap/tidb-lightning/blob/ccf7f3c3e5a80875a3a21b9b02e2657e7748a560/tests/checkpoint_chunks/run.sh#L76-L78

lance6716:
Ci fail:

`CREATE TABLE IF NOT EXISTS `double` (`ref` INT NOT NULL,`pk` DOUBLE(16) NOT NULL PRIMARY KEY);`

create table failed: Error 1064: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use 🤔

lightning/restore/restore.go
```diff
@@ -630,6 +616,16 @@ func (rc *RestoreController) runPeriodicActions(ctx context.Context, stop <-chan
...
 			)
+		case <-glueProgressTicker.C:
+			stage := glue.StageWriting
+			estimated := metric.ReadCounter(metric.ChunkCounter.WithLabelValues(metric.ChunkStateEstimated))
+			finished := metric.ReadCounter(metric.ChunkCounter.WithLabelValues(metric.ChunkStateFinished))
+			if finished >= estimated {
+				finished = estimated
+				stage = glue.StagePostProcessing
+			}
+			rc.tidbGlue.Record("ProgressPermillage", uint64(estimated/finished*1000))
```
kennytm, line 627: can we record the estimated and finished separately. the percent can be calculated by the glue impl.

kennytm (commented):
rest LGTM

lightning/restore/restore.go
lance6716: changed in https://github.com/pingcap/tidb-lightning/pull/456/commits/e880ad1d1475596cd4e021b2b9cac5101722a400

glorv (approved):
LGTM

PR #459, opened 2020-11-09
config: filter out all system schemas by default

kennytm:

If the data source contains system schemas, they may be accidentally get imported and corrupt the cluster.

Explicitly filter out system schemas (`mysql.*`, etc.) like pingcap/dumpling#194

### Check List 

Tests 

 - Unit test

Side effects

Related changes

# Release notes

* System schemas like `mysql.*` are excluded by default. Use `-f '*.*'` to explicitly include them.


overvenus (approved):
LGTM

glorv (approved):
LGTM

PR #471, opened 2020-11-11
config: allow four byte-size config to be specified using human-readable units ("100 GiB")

kennytm:

Several config specifies a large byte size, which is not very readable (and some DBAs are still unfamiliar with the underscore-as-number-separator TOML feature).

Using [`docker/go-units`](https://pkg.go.dev/github.com/docker/go-units) (already used by Dumpling and PD), we support parsing human-readable byte size for the following config:

* tikv-importer.region-split-size (100_663_296 → '96 MiB')
* mydumper.read-block-size (65536 → '64 KiB')
* mydumper.batch-size (107_374_182_400 → '100 GiB')
* mydumper.max-region-size (268_435_456 → '256 MiB')

Byte size can be specified using integers (`65536` or `65_536`, backward-compatible with existing config), floating point (`6.5536e+4`), or a string involving byte units (`'64k'`, `'64 K'`, `'64KiB'`, `'64 kb'`, `'0.0625 MB'`, all equivalent).

(This PR is submitted mainly because we expect that disk-quota is also going to need some huge numbers in a custom config, similar to TiKV's `ReadableSize` struct. Interestingly no such thing exists in TiDB except in dealing with INSPECTION_RULES.)

### Check List 

Tests 

 - Unit test

Side effects

Related changes

# Release notes

* The 4 configuration specified above can now accept human-readable format in the form `"2.5 GiB"`.

### DIFF
--- a/lightning/config/bytesize.go
+++ b/lightning/config/bytesize.go
@@ -1,0 +1,44 @@
+// Copyright 2020 PingCAP, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package config
+
+import (
+	"encoding/json"
+
+	"github.com/docker/go-units"
+)
+
+// ByteSize is an alias of int64 which accepts human-friendly strings like
+// '10G' when read from TOML.
+type ByteSize int64
+
+// UnmarshalText implements encoding.TextUnmarshaler
+func (size *ByteSize) UnmarshalText(b []byte) error {
+	res, err := units.RAMInBytes(string(b))
+	if err != nil {
+		return err
+	}
+	*size = ByteSize(res)
+	return nil
+}
+
+// UnmarshalJSON implements json.Unmarshaler (for testing)
+func (size *ByteSize) UnmarshalJSON(b []byte) error {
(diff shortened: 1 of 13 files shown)


lightning/restore/checksum_test.go
```diff
@@ -271,16 +271,17 @@ func (s *checksumSuite) TestGcTTLManagerSingle(c *C) {
...
 	c.Assert(err, IsNil)
 
-	time.Sleep(6 * time.Second)
+	time.Sleep(6*time.Second + 10*time.Millisecond)
 
-	// after 11 seconds, must at least update 5 times
+	// after 6 seconds, must at least update 5 times
 	val := atomic.LoadInt32(&pdClient.count)
-	c.Assert(val >= 5, IsTrue)
+	c.Assert(val, Ge, 5)
```
glorv, line 278: This should be `GreaterEqual` ?

lightning/config/bytesize.go
```diff
@@ -0,0 +1,44 @@
...
+	res, err := units.RAMInBytes(string(b))
+	if err != nil {
+		return err
+	}
+	*size = ByteSize(res)
+	return nil
+}
+
+// UnmarshalJSON implements json.Unmarshaler (for testing)
+func (size *ByteSize) UnmarshalJSON(b []byte) error {
```
glorv, line 37: shall we also implement `MarshalText`and `MarshalJSON`

kennytm: Seems no need to implement the Marshal methods as the unit test passes without them.

lightning/restore/checksum_test.go
kennytm: thanks, fixed.

glorv (approved):
LGTM

overvenus:
LGTM

PR #476, opened 2020-11-16
backend/local: more robust range retry strategy

glorv:
Add better range retry strategy to reduce the probability of `Epoch Not Match` error and make lightning run faster if the range conflict is very heavy.

- Remove the parallel write&ingest code. Since we already do concurrency at a higher level, this can make the retry logic clearer and easier.
- If a kv range is written to multi regions, only retry the failed regions.
- Only increase the retry counter if one retry has no progress at all

Benchmark result:

The benchmark was running  in the following two dataset:
1. dataset1. randomly generated int primary key. (50GB)
2. dataset2. randomly generated int primary key and 2 extra secondary keys. (126G)

lighting config:
```
[lightning]
 table-concurrency = 8

[mydumper]
batch-size = "1GiB"
```

| DateSet | branch  | EpochNotMatch errors count | cost time | 
| --- | --- | --- | --- |
| dataset1 | master | 1949 | 25m23s |
| dataset1 | retry-range | 196 | 6m38s |
| dataset2 | master | 4130 | 2h14m |
| dataset2 | retry-range | 298 | 50m15s |

### Check List 

Tests 

 - Unit test
 - Integration test
 - Manual test
 - No code

Side effects

Related changes


kennytm (commented):
LGTM

lightning/backend/local.go
```diff
@@ -892,65 +880,39 @@ WriteAndIngest:
...
-				if err1 != nil {
-					err = err1
-					log.L().Warn("should retry this range", zap.Int("retry", retry), log.ShortError(err))
-					shouldRetry = true
+			w := local.ingestConcurrency.Apply()
+			var rg *Range
+			rg, err = local.writeAndIngestPairs(ctx, engineFile, region, pairStart, end)
+			local.ingestConcurrency.Recycle(w)
+			if err != nil {
+				regionRange := intersectRange(region.Region, Range{start: pairStart, end: end})
```
lance6716, line 899: could we only decode and compare two "start" 🤔

lance6716:
/run-all-tests

[2020-11-25T11:24:29.480Z] ERROR 2013 (HY000): Lost connection to MySQL server at 'reading initial communication packet', system error: 104

lightning/backend/local.go
glorv: done. PTAL, again

lance6716 (approved):
LGTM

PR #477, opened 2020-11-17
glue: add GlueCheckpointDB and remove external TiDB usage

lance6716:
complete glue to enable integrated into TiDB

- add a glue implemented CheckpointDB, which is not elegant since it repeated many MySQLCheckpointDB code. Will refine later or in this PR.

- resome some usage of `TiDBManager`. now `NewTiDBManager` is only called in `tidb-lightning-ctl`

- remove usage of TiDB status port through TLS. now TLS is used only supposed to connect PD/TiKV

### Check List 

Tests 

 - pass original test

Side effects

 - Increased code complexity

Related changes

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,6 +48,9 @@ default: clean lightning lightning-ctl checksuccess
 prepare:
 	$(PREPARE_MOD)
 
+finish-prepare:
+	$(FINISH_MOD)
+
 clean:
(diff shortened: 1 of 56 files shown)


lance6716:
will reduce commit message

kennytm:
You could just `git commit --amend` and then `git push --force`, no need to create a new PR.

lance6716:
> You could just `git commit --amend` and then `git push --force`, no need to create a new PR.

Oh I should refresh this page. I force pushed and saw old commits still here 😂

PR #478, opened 2020-11-17
glue: add GlueCheckpointDB and remove external TiDB usage

lance6716:
complete glue to enable integrated into TiDB

- add a glue implemented CheckpointDB, which is not elegant since it repeated many MySQLCheckpointDB code. Will refine later or in this PR.
- resome some usage of `TiDBManager`. now `NewTiDBManager` is only called in `tidb-lightning-ctl`
- remove usage of TiDB status port through TLS. now TLS is used only supposed to connect PD/TiKV
- move database name from `USE db` to CREATE TABLE statements, because we may not rely on idle `(sql.DB).ExecContext` use same connection to execute sequentially SQLs, and in future we may support concurently create tables
- remove `taskID` and replace it with `cfg.TaskID`, because `taskID` didn't have another value

### Check List 

Tests 

 - manually test integrated into TiDB and import a sample data from integration test

Side effects

 - Increased code complexity

Related changes

### DIFF
--- a/lightning/checkpoints/checkpoints.go
+++ b/lightning/checkpoints/checkpoints.go
@@ -67,6 +67,115 @@ const (
 	CheckpointTableNameChunk  = "chunk_v5"
 )
 
+const (
+	// shared by MySQLCheckpointsDB and GlueCheckpointsDB
+	CreateDBTemplate        = "CREATE DATABASE IF NOT EXISTS %s;"
+	CreateTaskTableTemplate = `
+		CREATE TABLE IF NOT EXISTS %s.%s (
+			id tinyint(1) PRIMARY KEY,
+			task_id bigint NOT NULL,
+			source_dir varchar(256) NOT NULL,
+			backend varchar(16) NOT NULL,
+			importer_addr varchar(256),
+			tidb_host varchar(128) NOT NULL,
+			tidb_port int NOT NULL,
+			pd_addr varchar(128) NOT NULL,
+			sorted_kv_dir varchar(256) NOT NULL,
+			lightning_ver varchar(48) NOT NULL
+		);`
+	CreateTableTableTemplate = `
+		CREATE TABLE IF NOT EXISTS %s.%s (
+			task_id bigint NOT NULL,
+			table_name varchar(261) NOT NULL PRIMARY KEY,
+			hash binary(32) NOT NULL,
+			status tinyint unsigned DEFAULT 30,
+			alloc_base bigint NOT NULL DEFAULT 0,
+			table_id bigint NOT NULL DEFAULT 0,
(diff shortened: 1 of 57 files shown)


lance6716:
To design an interface to implemented by both TiDB's data structure (most work) and database/sql, and make `MySQLCheckpointDB` reusable, I found those works are needed:

- to support `(RowInterface).Scan(dst ...interface{})`, in other words to convert a row from TiDB's result set to golang's variables accroding to their types, there's ~260 lines needed https://github.com/golang/go/blob/869e2957b9f66021581b839cadce6cb48ad46114/src/database/sql/convert.go#L219 
- to support `Exec(query string, args ...interface{})`, in other words to interpolate arguments to SQL string, ~100 lines needed https://github.com/go-sql-driver/mysql/blob/3b935426341bc5d229eafd936e4f4240da027ccd/connection.go#L198

lines are significently less than `glue_checkpoint.go`. 

I guess we must repeat some code of `MySQLCheckpointDB` (currently) or repeat some code of go's SQL driver to implement glue checkpoint. Should I switch to another way? @kennytm @glorv @overvenus

overvenus:
For now, I don't think we need to implement 'Scan' and 'Exec'. We can go on with the current implementation.

lightning/checkpoints/checkpoints.go
```diff
@@ -67,6 +67,115 @@ const (
...
+		WHERE (table_name, engine_id, path, offset) = (?, ?, ?, ?);`
+	UpdateTableRebaseTemplate = `
+		UPDATE %s.%s SET alloc_base = GREATEST(?, alloc_base) WHERE table_name = ?;`
+	UpdateTableStatusTemplate = `
+		UPDATE %s.%s SET status = ? WHERE table_name = ?;`
+	UpdateEngineTemplate = `
+		UPDATE %s.%s SET status = ? WHERE (table_name, engine_id) = (?, ?);`
+	DeleteChunkTemplate  = "DELETE FROM %s.%s WHERE table_name = ?;"
+	DeleteEngineTemplate = "DELETE FROM %s.%s WHERE table_name = ?;"
+	DeleteTableTemplate  = "DELETE FROM %s.%s WHERE table_name = ?;"
```
kennytm, line 176: can we use a single constant for all three? 😅

kennytm (commented):
Rest LGTM

lance6716:
Add new change:

- move database name from `USE db` to CREATE TABLE statements, because we may not rely on idle `(sql.DB).ExecContext` use same connection to execute sequentially SQLs, and in future we may support concurently create tables

PTAL @kennytm @glorv @overvenus

kennytm (commented):
LGTM

glorv (approved):
LGTM

Issue #446, opened 2020-11-04
Hope to specify a disk size to check point


shuijing198799:
## Feature Request

**Is your feature request related to a problem? Please describe:**

The lightning need a volume to save check point file, It’s hard to predict the size of this disk， So we must prepare as large a disk as possible to install these temporary files,  For example , If I need to restore 2T data, we have to prepare a 2T volume for lightning, It is a bad experience to use on the cloud。

**Describe the feature you'd like:**

We want to specify the volume size and the checkpoint will exceed this size in the lightning process.

**Describe alternatives you've considered:**

none

**Teachability, Documentation, Adoption, Optimization:**

kennytm:
by "check point file" you mean those "SST files" in the local mode?

shuijing198799:
> by "check point file" you mean those "SST files" in the local backend?

yes。

kennytm:
Seems we can use https://pkg.go.dev/github.com/cockroachdb/pebble#DB.EstimateDiskUsage to fetch the disk usage.

# Abstract 

Periodically, before a `WriteRows`, check every engine's total estimated disk usage. If the total disk usage exceeds the "(soft) disk quota", we block the write to the largest engines until the remaining total is less than the quota, and flush the blocked engines' content into TiKV. The engine UUIDs are reused.

This will cause subsequent imports to suffer from range overlapping, which we have to accept as trade-off.

# Checkpoint validity

The flushing design must be compatible with checkpoints, that is no data will be lost if we Ctrl+C → resume in the middle of a process. Checkpoints may be earlier than the actual progress, so some data (process) duplication should be acceptable and ignored.

Now let's consider the flush process:

1. ... parallel WriteRows ...
2. detected quota overflow, start emergency ingest to TiKV
3. CloseEngine()
    1. Flush()
    2. saveEngineMeta()
4. ImportEngine()
    1. readAndSplitIntoRange()
    2. loop:
        * SplitAndScatterRegionByRanges()
        * WriteAndIngestByRanges()
5. Reset engine to empty
6. ... parallel WriteRows ...

Let's consider what happens regarding the place of interruption (I) and actual saved checkpoint (C):

## Case I=3, C<3

Currently, with Local backend, a checkpoint is flushed only when the entire engine is written because Flush() is expensive (https://github.com/pingcap/tidb-lightning/pull/326#issuecomment-638725946). So the end of step 3 is a good point to save the checkpoint. 

[...]

PR #493, opened 2020-11-26
backend: implement disk quota

kennytm:

Implement #446.

### Check List 

Tests 

 - Unit test
 - Integration test

Side effects

 - Increased code complexity

Related changes


lightning/backend/backend.go
```diff
@@ -212,6 +246,61 @@ func (be Backend) FetchRemoteTableModels(ctx context.Context, schemaName string)
...
+	largeEngines []uuid.UUID,
+	inProgressLargeEngines int,
+	totalDiskSize int64,
+	totalMemSize int64,
+) {
+	sizes := be.abstract.EngineFileSizes()
+	sort.Slice(sizes, func(i, j int) bool {
+		a, b := &sizes[i], &sizes[j]
+		if a.IsImporting != b.IsImporting {
+			return a.IsImporting
```
Little-Wallace, line 266: What does it mean? Do you want to sort all importing engine  before other engines?

kennytm: Yes. We can't flush engines that are already importing, but their size do count towards the disk quota.

kennytm:
The *test case* depends on #505 in order to quickly generate huge SSTs without checking in a 1 GB database or introduce dbgen.

lightning/backend/local.go
```diff
@@ -95,19 +98,32 @@ type Range struct {
...
 
 type LocalFile struct {
 	localFileMeta
 	db   *pebble.DB
 	Uuid uuid.UUID
+
+	// importSemaphore ensures the engine can only be imported by a single
+	// thread. We use a semaphore instead of a sync.Mutex because we need the
+	// TryAcquire method.
+	importSemaphore *semaphore.Weighted
```
lance6716, line 116: is it faster than select chan with default branch 🤔

lightning/backend/local.go
```diff
@@ -1002,44 +1082,45 @@ loopWrite:
...
-	for i := 0; i < len(ranges); i++ {
-		// wait for all sub tasks finish to avoid panic. if we return on the first error,
-		// the outer tasks may close the pebble db but some sub tasks still read from the db
-		e := <-errCh
-		if e != nil && err == nil {
-			err = e
-		}
-	}
-	return err
+	// wait for all sub tasks finish to avoid panic. if we return on the first error,
```
lance6716, line 1120: could we use another way to support both return early and don't panic, such as recovering, peeking context? I'm not sure if these calls to pd and TiKV is slow

kennytm: dunno. but this part is basically just a refactor of the existing code.

lightning/backend/local.go
kennytm: actually we don't need TryAcquire now 😅, so switching back to regular mutex.

lightning/restore/restore.go
```diff
@@ -1742,45 +1829,54 @@ func (cr *chunkRestore) deliverLoop(
...
-		if err = dataEngine.WriteRows(ctx, columns, dataKVs); err != nil {
-			deliverLogger.Error("write to data engine failed", log.ShortError(err))
-			return
-		}
-		if err = indexEngine.WriteRows(ctx, columns, indexKVs); err != nil {
-			deliverLogger.Error("write to index engine failed", log.ShortError(err))
-			return
-		}
+		func() {
+			rc.diskQuotaLock.RLock()
```
lance6716, line 1838: in #446 
> we block the write to the **largest engines** until the remaining total is less than the quota

but here we block all engines. is this expected?

lightning/config/config.go
```diff
@@ -532,6 +559,40 @@ func (cfg *Config) Adjust(ctx context.Context) error {
...
+			writeAmount := uint64(cfg.App.RegionConcurrency) * uint64(cfg.Cron.CheckDiskQuota.Milliseconds())
+			reservedSize := enginesCount*autoDiskQuotaLocalReservedSize + writeAmount*autoDiskQuotaLocalReservedSpeed
+
+			storageSize, err := common.GetStorageSize(storageSizeDir)
+			if err != nil {
+				return err
+			}
+			if storageSize.Available <= reservedSize {
+				return errors.Errorf(
+					"insufficient disk free space on `%s` (only %s, expecting >%s)",
```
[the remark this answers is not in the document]
kennytm: Reducing CheckDiskQuota duration typically has less effect than reducing the two concurrencies. As described in the comment on `autoDiskQuotaLocalReservedSize` and `autoDiskQuotaLocalReservedSpeed`, their contributions are by default 4.0 GiB (index+table conc = 8) and 2.3 GiB (region conc = 40) respectively. Even if we have 100 region conc the contribution is still just 5.8 GiB. And I'm not quite worried about having a 10 GiB buffer for the default case when the data source and disk size is typically an order of magnitude larger.

Anyway expanded the error message asking them to define `tikv-importer.disk-quota` as an alternative (which will skip this check).

lightning/config/config_test.go
```diff
@@ -144,9 +145,9 @@ func (s *configTestSuite) TestAdjustFileRoutePath(c *C) {
 	rule := &config.FileRouteRule{Path: invalidPath, Type: "sql", Schema: "test", Table: "tbl"}
 	cfg.Mydumper.FileRouters = []*config.FileRouteRule{rule}
 	err := cfg.Adjust(ctx)
-	c.Assert(err, ErrorMatches, fmt.Sprintf("file route path '%s' is not in source dir '%s'", invalidPath, tmpDir))
+	c.Assert(err, ErrorMatches, fmt.Sprintf("\\Qfile route path '%s' is not in source dir '%s'\\E", invalidPath, tmpDir))
```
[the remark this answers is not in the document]
lance6716: seems regular expression grammar, means matching exactly start/end

tests/disk_quota/data/disk_quota.t-schema.sql
```diff
@@ -0,0 +1,12 @@
+create table t (
+    id int not null primary key,
+
+    -- each stored generated column occupy about 150 KB of data, so we are 750 KB per row.
+    -- without disk quota the engine size will be 750 KB * 2000 rows = 1.5 GB ≈ 1.4 GiB.
+    -- (FIXME: making the KV size too large may crash PD?)
```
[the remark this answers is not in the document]
kennytm: No stack trace. But I was testing on the laptop so it's likely just environment issue (lack of memory).

lightning/config/config_test.go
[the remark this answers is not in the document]
kennytm: see https://pkg.go.dev/regexp/syntax for the complete regexp grammar. `\Q...\E` is listed under "Escape sequences".

lightning/config/config.go
```diff
@@ -532,6 +559,40 @@ func (cfg *Config) Adjust(ctx context.Context) error {
...
+			storageSizeDir = filepath.Dir(storageSizeDir)
+		case err == nil:
+			if !sortedKVDirInfo.IsDir() {
+				return errors.Errorf("tikv-importer.sorted-kv-dir ('%s') is not a directory", storageSizeDir)
+			}
+		default:
+			return errors.Annotate(err, "invalid tikv-importer.sorted-kv-dir")
+		}
+
+		if cfg.TikvImporter.DiskQuota == 0 {
```
glorv, line 578: I think we need to also take the source files' total size into consideration.

lightning/restore/restore.go
```diff
@@ -1321,6 +1334,80 @@ func (rc *RestoreController) switchTiKVMode(ctx context.Context, mode sstpb.Swit
...
 
+func (rc *RestoreController) enforceDiskQuota(ctx context.Context) {
+	if !atomic.CompareAndSwapInt32(&rc.diskQuotaState, diskQuotaStateIdle, diskQuotaStateChecking) {
+		// do not run multiple the disk quota check / import simultaneously.
+		// (we execute the lock check in background to avoid blocking the cron thread)
+		return
+	}
+
+	go func() {
+		rc.diskQuotaLock.Lock()
```
glorv, line 1345: Is it better to use an RLock() to check the disk quota? if the quota is exceeded, then change the RLock() to Lock() so we can avoid block all the writers here.

lightning/restore/restore.go
```diff
@@ -1742,45 +1829,54 @@ func (cr *chunkRestore) deliverLoop(
 			}
 		}
 
-		// Write KVs into the engine
-		start := time.Now()
+		// in local mode, we should write these checkpoint after engine flushed
+		// we are allowed to save checkpoint when the disk quota state moved to "importing"
+		// since all engines are flushed.
+		canSaveCheckpoint := !rc.isLocalBackend() || atomic.LoadInt32(&rc.diskQuotaState) == diskQuotaStateImporting
```
glorv, line 1835: If the current state is `diskQuotaStateImporting`, then the unsafe import and reset routine should hold the diskQuotaLock, then the RLock here should be blocked.  And after all the unsafe import tasks are finished sometime later, the disk quote state should be switched to normal. Then the write rows call will write fresh data to local storage without being flushed, thus the save chunk checkpoint is not safe.

glorv: I think it's better just call `saveCheckpoint(rc, t, engineID, cr.chunk)` if current status is diskQuotaStateImporting before call WriteRows below

lightning/backend/local.go
```diff
@@ -1078,6 +1130,9 @@ func (local *local) ImportEngine(ctx context.Context, engineUUID uuid.UUID) erro
 		return nil
 	}
 
+	atomic.StoreInt32(&lf.isImporting, 1)
```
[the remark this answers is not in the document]
kennytm: Not really necessary since this is executed inside a lock 😅. we use atomic mainly for the `isImporting()` function.

lightning/restore/restore.go
kennytm: 🤔 we could move the write-lock next to `logger.Warn("disk quota exceeded")`, though that means the writer are still writing while we are checking disk space (doesn't really matter due to the soft-limit designation). acquiring a read-lock here is useless since it doesn't block anything.

lightning/restore/restore.go
```diff
@@ -1332,6 +1345,91 @@ func (rc *RestoreController) switchTiKVMode(ctx context.Context, mode sstpb.Swit
...
+			// if any engine failed to import, we just try again next time, since the data are still intact.
+			atomic.StoreInt32(&rc.diskQuotaState, diskQuotaStateImporting)
+			task := logger.Begin(zap.WarnLevel, "importing large engines for disk quota")
+			var importErr error
+			for _, engine := range largeEngines {
+				if err := rc.backend.UnsafeImportAndReset(ctx, engine); err != nil {
+					importErr = multierr.Append(importErr, err)
+				}
+			}
+			task.End(zap.ErrorLevel, importErr)
```
glorv, line 1427: What if some or all engines imported failed here, e.g. by splitting region failures or one or more tikv instances are down. In the current implementation, this function will return with an error, and all write will then resume while the disk is nearly full before the next `checkDiskQuota` is scheduled. I'm afraid in some conditions, the pebble routine will be stuck at flush when the disk is full as we have met.

lightning/backend/local.go
```diff
@@ -1156,11 +1247,30 @@ func (local *local) ImportEngine(ctx context.Context, engineUUID uuid.UUID) erro
 	return nil
 }
 
+func (local *local) ResetEngine(ctx context.Context, engineUUID uuid.UUID) error {
+	// the only way to reset the engine + reclaim the space is to delete and reopen it 🤷
+	if err := local.CleanupEngine(ctx, engineUUID); err != nil {
+		return err
+	}
+	if err := local.OpenEngine(ctx, engineUUID); err != nil {
```
glorv, line 1255: I think here, it's better to only clean up the engine, and left the lazy open engine operation to write rows. In this way we can avoid the corner case that the extra open engine may never be visited again.
Example(if check disk quota and the normal engine write finish at the same time, and this engine is big enough to trigger the unsafe import):
```
disk-quote-routine                         normal-engine-routine
1. lock-and-import
                                            2. lock-and-import
                                            3. clean-up-engine
4. cleanup-and-reopen
```
In this timeline, after action 4, the reopened engine is never visited again, and it's data files won't be clean up either.

glorv (commented):
rest LGTM

lightning/config/config_test.go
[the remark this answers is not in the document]
overvenus: I see, thanks!

lightning/restore/restore.go
kennytm: 🤔 no good idea yet.

things we know won't work:

* retry import on error - we already retried internally, more retry won't help
* use a state to record if the engine has been flushed to avoid double-flushing - won't work because after the mutex is unlocked, some more rows will be written to make the engine dirty again
* just skip 1 minute (replace the `return` next line by `continue`) - almost same as retry, and we still double-flush, even without double-flush we can still be stuck

lightning/backend/local.go
kennytm: changed to do nothing at step 4 if the engine is already gone.

lightning/backend/local.go
```diff
@@ -1159,24 +1251,50 @@ func (local *local) ImportEngine(ctx context.Context, engineUUID uuid.UUID) erro
...
+
+	localEngine := engineFile.(*LocalFile)
+	localEngine.lock()
+	defer localEngine.unlock()
+
+	// since closing the engine causes all subsequent operations on it panic,
+	// we make sure to delete it from the engine map before calling Close().
+	// (note that Close() returning error does _not_ mean the pebble DB
+	// remains open/usable.)
+	local.engines.Delete(engineUUID)
```
lance6716, line 1281: could this `Delete` happen at same when index engine is flushing in 
https://github.com/pingcap/tidb-lightning/blob/463e7289fcbc412a49e50b307e3b33d30da7f218/lightning/restore/restore.go#L1160
? so flush may access a closed engine. we might need a lock to protect `local.engines`, or extend `LocalFile.importMutex` to protect this close action.

lightning/backend/local.go
```diff
@@ -1162,24 +1272,50 @@ func (local *local) ImportEngine(ctx context.Context, engineUUID uuid.UUID) erro
...
+
+	localEngine := engineFile.(*LocalFile)
+	localEngine.lock(importMutexStateClose)
+	defer localEngine.unlock()
+
+	// since closing the engine causes all subsequent operations on it panic,
+	// we make sure to delete it from the engine map before calling Close().
+	// (note that Close() returning error does _not_ mean the pebble DB
+	// remains open/usable.)
+	local.engines.Delete(engineUUID)
```
lance6716, line 1302: seems `pebble.DB` of engines are flushed when `CloseEngine`, while `CloseEngine` didn't check the lock of `LocalFile` to prevent `pebble.DB.Close` in `cleanupEngineIfExists`

lightning/backend/local.go
```diff
@@ -1112,6 +1218,10 @@ func (local *local) ImportEngine(ctx context.Context, engineUUID uuid.UUID) erro
 	}
 
 	lf := engineFile.(*LocalFile)
+
+	lf.lock(importMutexStateImport)
+	defer lf.unlock()
+
```
lance6716, line 1224: we might reload engine file from `local.engines` here? or we might always lock the **UUID** (maybe a `map[UUID]mutex` like `local.engines`, to ge rid of getting the file -> checking the lock -> getting file again) before accessing the `LocalFile`?

And not sure `ResetEngine` should hold that UUID lock all through the function

lance6716: disk quota routine is concurrent with normal open-write-close-import routine, and engines are flush/close/import/reopen in two rontine but at same UUID, that may make things easier to understood 😵 

and maybe UUID level lock instead of `rc.diskQuotaLock` could avoid blocking all engines' writing

glorv: But the requirement is to lock all engines to avoid any further writing to disk since the disk is near full..

lightning/backend/local.go
glorv: Agree, seems `LocalFile.Close` should also acquire the mutex first

lightning/backend/local.go
lance6716: I was recalling https://github.com/pingcap/tidb-lightning/pull/493#discussion_r539021798

if the "soft limit" works well, we could only block the `largeEngines` returned in `CheckDiskQuota`

lance6716:
since `rc.diskQuotaLock` only block Write in "Open -> Write -> (Flush of index engine -> ) Close -> Import", for Flush, Close, Import stage we might also acquire UUID lock during the stage to avoid concurrent access to a engine with disk quota goroutine (currently Close does not respect the lock. Also we should note that Close has a functionality of recovery mode).

And in disk quota goroutine during the locking scope of `rc.diskQuotaLock` there's "FlushAll -> for engine UnsafeImportAndReset". the OpenEngine of ResetEngine of UnsafeImportAndReset does not respect the lock. Besides that, IMO ResetEngine should keep the old UUID lock instead of deleting and adding a new one, otherwise normal engine routine may go wrong.

lightning/restore/restore.go
Little-Wallace: Shall we also block `encodeLoop` ? If `deliverLoop` blocks while `encodeLoop` continues processing data, the data may stay in channel and may cause "out of memory".

lance6716: the channel is limit to `maxKVQueueSize` capacity and so `encodeLoop` will be blocked after full 🤔

lightning/restore/restore.go
Little-Wallace: It seems that we do not change the checkpoint after flushing engines? So why we must flush them before importing?

kennytm:
(blocked by #543)

lightning/restore/restore.go
Little-Wallace: If `deliverLoop` was blocked in "#1923"(`rc.diskQuotaLock.RLock()`), it would not save checkpoint, would it?

overvenus:
(#543 is merged)

overvenus (commented):
LGTM

overvenus:
Please resolve conflicts.

glorv:
Please fix the tests

lightning/backend/local.go
```diff
@@ -423,16 +456,62 @@ func (local *local) Close() {
...
+	var flushFinishedWg sync.WaitGroup
+	var flushedEngines []*LocalFile
+
+	// first, (asynchronously) flush all opened engines
+	local.engines.Range(func(k, v interface{}) bool {
+		engineFile := v.(*LocalFile)
+		if engineFile.isLocked() {
+			// ignore engines which have been importing, flushing, or closing.
+			return true
```
3pointer, line 487: At this point
All OpenEngine is blocked due to [DiskQuotaLock](https://github.com/pingcap/tidb-lightning/pull/493/files#diff-ca1e86afc37a9b71183fb70c7fd2d65ea68a24a60e55a664d2f379be63fc7f9dR1727), so there is no new opened engine.
All exists engineFile is locked to flush state due to [L476](https://github.com/pingcap/tidb-lightning/pull/493/files#diff-3f1c53e9d349649b588b6da73533591d2f35d992bbafe5239cc3cc781f71e7e8R476), so It's always return true.
It seems the code will never reach the `AsyncFlush` logic. 🤔️

lightning/backend/local.go
```diff
@@ -1609,186 +1744,307 @@ func (s *sizeProperties) iter(f func(p *rangeProperty) bool) {
...
+func (local *local) EngineFileSizes() (res []EngineFileSize) {
+	local.engines.Range(func(k, v interface{}) bool {
+		engine := v.(*LocalFile)
+		res = append(res, engine.getEngineFileSize())
+		return true
+	})
+	return
+}
+
 type LocalWriter struct {
```
kennytm, line 1757: PTAL @Little-Wallace, the LocalWriter is basically entirely rewritten, there may be some performance implication of the original code that I may have missed. Thanks.

lightning/restore/restore.go
glorv: yes, the current implementation can not handle this if the previous check isImporting is false and the  disk-quote loop holds the lock and do importing.

glorv (commented):
rest LGTM

lightning/backend/local.go
```diff
@@ -406,11 +484,13 @@ func (local *local) getGrpcConnLocked(ctx context.Context, storeID uint64) (*grp
 
 // Close the importer connection.
```
lance6716, line 485: found an outdated comment

kennytm: 😅

lightning/backend/local.go
```diff
@@ -406,11 +484,13 @@ func (local *local) getGrpcConnLocked(ctx context.Context, storeID uint64) (*grp
...
 func (local *local) Close() {
-	local.engines.Range(func(k, v interface{}) bool {
-		v.(*LocalFile).Close()
-		return true
-	})
+	allEngines := local.lockAllEnginesUnless(importMutexStateClose, 0)
```
lance6716, line 487: some functions like `AppendRows` didn't check context cancel, I'm afraid that some parent goroutine exits because of context cancel (like line 2429 in restore.go) while the child goroutine is still running, which may cause a "false" 0 in time slice between unlocking and locking.

lightning/backend/local.go
```diff
@@ -177,17 +187,81 @@ func (e *LocalFile) getSizeProperties() (*sizeProperties, error) {
...
+	e.isImportingAtomic.Store(uint32(state))
+}
+
+func (e *LocalFile) lockUnless(newState, ignoreStateMask importMutexState) bool {
+	for {
+		curState := e.isImportingAtomic.Load()
+		if curState&uint32(ignoreStateMask) != 0 {
+			return false
+		}
+		if e.isImportingAtomic.CAS(curState, uint32(newState)) {
```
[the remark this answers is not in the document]
kennytm: just changed to call `.lock()` directly.

kennytm (commented):
~~test~~

lightning/backend/local.go
```diff
@@ -177,17 +186,122 @@ func (e *LocalFile) getSizeProperties() (*sizeProperties, error) {
...
+			case err := <-replyErrCh:
+				return err
+			}
+		})
+		return true
+	})
+	return eg.Wait()
+}
+
+func (e *LocalFile) flushEngineUnlocked(ctx context.Context) error {
```
lance6716, line 264: not understand what "Unlocked" means. This method is called with engine file locked

lightning/backend/local.go
```diff
@@ -521,33 +631,29 @@ func (local *local) CloseEngine(ctx context.Context, engineUUID uuid.UUID) error
...
 	engineFile := engine.(*LocalFile)
-	err := engineFile.db.Flush()
-	if err != nil {
+	engineFile.lock(importMutexStateFlush)
+	defer engineFile.unlock()
+	if err := engineFile.saveEngineMeta(); err != nil {
 		return err
 	}
-	return local.saveEngineMeta(engineFile)
+	return engineFile.db.Flush()
```
lance6716, line 656: seems we didn't flush localWriter

lightning/backend/importer.go
```diff
@@ -53,6 +53,9 @@ type importer struct {
 	tls    *common.TLS
 
 	mutationPool sync.Pool
+	// importLock is a backend-global lock to ensure only one ImportEngine() is
+	// running at a time slice.
+	importLock sync.Mutex
```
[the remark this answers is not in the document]
glorv: remove for what?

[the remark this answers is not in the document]
lance6716: we have make sure import is not concurrent, such as

https://github.com/pingcap/tidb-lightning/blob/90dbfe3c143a3448c0dbd74845e92089d813cccb/lightning/restore/restore.go#L1267-L1270

lightning/backend/local.go
```diff
@@ -1609,186 +1730,307 @@ func (s *sizeProperties) iter(f func(p *rangeProperty) bool) {
...
+// kvMemCache is an array of KV pairs. It also keep tracks of the total KV size and whether the array is already sorted.
+type kvMemCache struct {
+	kvs       []common.KvPair
+	totalSize int64
+	notSorted bool // record "not sorted" instead of "sorted" so that the zero value is correct.
+}
+
+// append more KV pairs to the kvMemCache.
+func (m *kvMemCache) append(kvs []common.KvPair) {
+	if !m.notSorted {
```
Little-Wallace, line 2003: Even if the kvMemCache has been sorted, we shall still check the order of `kvs`. The array `kvs` may be unordered and the first key of `kvs` may be greater than the lastkey of `m.kvs`

lightning/backend/local.go
kennytm: it means this function itself has no locking. any better name 🤗

lightning/backend/local.go
kennytm: the loop inside does check the order of `kvs`.

```go
		var lastKey []byte
		if len(m.kvs) > 0 {
			lastKey = m.kvs[len(m.kvs)-1].Key
		}
		for _, kv := range kvs {
			if bytes.Compare(kv.Key, lastKey) <= 0 {    // <----
				m.notSorted = true
				break
			}
			lastKey = kv.Key   // <----
		}
```

lightning/backend/local.go
lance6716: no good idea 😂 maybe we could remove "Unlocked". I can't think out a name clearly indicating locking is prerequisite or locking is behaviour of the function. WDYT @glorv @overvenus @Little-Wallace

ti-srebot (approved):

PR #497, opened 2020-11-30
backend/local: check and return iter.Error when pebble is not valid

glorv:
Check pebble iter.Error when iter is not valid so if iter returns an error (such as caused by exceeding max open file limit), lightning won't omit the actual error

### Check List 

Tests 

 - Unit test
 - Integration test

Side effects

Related changes


lightning/backend/local.go
```diff
@@ -705,16 +709,24 @@ func (local *local) readAndSplitIntoRange(engineFile *LocalFile) ([]Range, error
 	iter := engineFile.db.NewIter(nil)
 	defer iter.Close()
 
+	iterError := func(e string) error {
+		err := iter.Error()
+		if err != nil {
+			return errors.Annotatef(err, e)
```
kennytm, line 715:
```suggestion
			return errors.Annotate(err, e)
```

lightning/backend/local.go
```diff
@@ -500,8 +500,12 @@ func (local *local) WriteToTiKV(
 	opt := &pebble.IterOptions{LowerBound: regionRange.start, UpperBound: regionRange.end}
 	iter := engineFile.db.NewIter(opt)
 	defer iter.Close()
+	iter.First()
+	if iter.Error() != nil {
+		return nil, nil, errors.Annotatef(iter.Error(), "failed to read the first key")
```
kennytm, line 505:
```suggestion
		return nil, nil, errors.Annotate(iter.Error(), "failed to read the first key")
```

lightning/backend/local.go
```diff
@@ -851,6 +863,9 @@ func (local *local) writeAndIngestByRange(
 	defer iter.Close()
 	// Needs seek to first because NewIter returns an iterator that is unpositioned
 	hasKey := iter.First()
+	if iter.Error() != nil {
+		return errors.Annotatef(iter.Error(), "failed to read the first key")
```
kennytm, line 867:
```suggestion
		return errors.Annotate(iter.Error(), "failed to read the first key")
```

kennytm (commented):
Rest LGTM

kennytm:
LGTM

kennytm:
/test

```
[2020-11-30T10:22:17.595Z] 
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0curl: (7) Failed connect to fileserver.pingcap.net:80; Connection refused
```

lightning/backend/local.go
```diff
@@ -500,8 +500,12 @@ func (local *local) WriteToTiKV(
 	opt := &pebble.IterOptions{LowerBound: regionRange.start, UpperBound: regionRange.end}
 	iter := engineFile.db.NewIter(opt)
 	defer iter.Close()
+	iter.First()
+	if iter.Error() != nil {
+		return nil, nil, errors.Annotate(iter.Error(), "failed to read the first key")
+	}
 
-	if !iter.First() {
+	if !iter.Valid() {
```
overvenus, line 508: What about `iter.Last()` at line 516?

lightning/backend/local.go
```diff
@@ -851,6 +863,9 @@ func (local *local) writeAndIngestByRange(
 	defer iter.Close()
 	// Needs seek to first because NewIter returns an iterator that is unpositioned
 	hasKey := iter.First()
+	if iter.Error() != nil {
+		return errors.Annotate(iter.Error(), "failed to read the first key")
+	}
```
overvenus, line 868: What about iter.Last() at line 877?

overvenus: What about this comment? @glorv

glorv: fixed

overvenus:
@glorv Please make sure this PR adds all necessary error checks for pebble usage.

glorv:
> @glorv Please make sure this PR adds all necessary error checks for pebble usage.

I have checked all the pebble iterators, there should be no remain.

lightning/backend/local.go
lance6716: do we need to check `Valid()` after `iter.Last()`, as we did for `iter.First()`?

glorv: yes,  see #880-881 @lance6716

lance6716: Do you mean `iter.Error()` here could cover `iter.Valid()`? For `iter.First()` why we need them both (maybe iterator will keep valid after normal operation on iterator 🤔 )

glorv: we only need to check iter.Error() != nil which  indicates that something goes wrong. iter.Error() != nil always means iter.Valid() is false. if iter.Error() == nil and iter.Valid() is false, this means the iterator iters to the end, this is ok to our logic.

lance6716:
LGTM

lance6716 (approved):

PR #501, opened 2020-12-01
backend/local: set pebble db max file limit

glorv:
Set the maximum open file limit for pebble to avoid open files exceeds the system open files limit

- set the estimated max open files limit to (sum of topN table files size) / tableConcurrency * 2
- check the estimated open files limit with a default min value of 1024. 
- Set pebble maxOpenFiles to the estimatedOpenFiles, L0CompactionThreshold and L0StopWritesThreshold to estimatedOpenFiles/2.

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Side effects

   - Since if there are more files in pebble than estimatedOpenFiles/2, pebble will start compacting L0 SST files and stop write, thus there may be performance regression in some extreme condition
 - Increased code complexity

Related changes

### DIFF
--- a/lightning/backend/local.go
+++ b/lightning/backend/local.go
@@ -77,7 +77,8 @@ const (
 	defaultPropSizeIndexDistance = 4 * 1024 * 1024 // 4MB
 	defaultPropKeysIndexDistance = 40 * 1024
 
-	minOpenFilesCount = 128
+	// the lower threshold of max open files for pebble db.
+	openFilesLowerThreshold = 128
 )
 
 var (
@@ -295,7 +296,7 @@ func NewLocalBackend(
 		tcpConcurrency:    rangeConcurrency,
 		batchWriteKVPairs: sendKVPairs,
 		checkpointEnabled: enableCheckpoint,
-		maxOpenFiles:      utils.MaxInt(maxOpenFiles, minOpenFilesCount),
+		maxOpenFiles:      utils.MaxInt(maxOpenFiles, openFilesLowerThreshold),
 	}
 	local.conns.conns = make(map[uint64]*connPool)
 	return MakeBackend(local), nil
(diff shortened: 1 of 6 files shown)


lightning/backend/local.go
```diff
@@ -291,6 +295,7 @@ func NewLocalBackend(
 		tcpConcurrency:    rangeConcurrency,
 		batchWriteKVPairs: sendKVPairs,
 		checkpointEnabled: enableCheckpoint,
+		maxOpenFiles:      utils.MaxInt(maxOpenFiles, minOpenFilesCount),
```
lance6716, line 298: `minOpenFilesCount` is confusing, maybe xxxLowerBound

lightning/backend/local.go
```diff
@@ -387,13 +392,14 @@ func (local *local) ShouldPostProcess() bool {
...
 	opt := &pebble.Options{
-		MemTableSize:             LocalMemoryTableSize,
-		MaxConcurrentCompactions: 16,
-		L0CompactionThreshold:    math.MaxInt32, // set to max try to disable compaction
-		L0StopWritesThreshold:    math.MaxInt32, // set to max try to disable compaction
-		MaxOpenFiles:             10000,
-		DisableWAL:               true,
-		ReadOnly:                 readOnly,
+		MemTableSize:                LocalMemoryTableSize,
+		MemTableStopWritesThreshold: 8,
```
lance6716, line 396: What's this option used for, avoid generating a big L0 file?

lightning/backend/local.go
```diff
@@ -387,13 +392,14 @@ func (local *local) ShouldPostProcess() bool {
...
+		MaxConcurrentCompactions:    16,
+		L0CompactionThreshold:       local.maxOpenFiles / 2, // set to max try to disable compaction
```
lance6716, line 398: outdated comments

lightning/backend/local.go
zhangjinpeng1987: agree

lightning/backend/local.go
glorv: has added a comment

overvenus:
LGTM

lightning/backend/local_unix.go
```diff
@@ -20,12 +20,29 @@ import (
...
+
+	"github.com/pingcap/tidb-lightning/lightning/log"
+)
+
+const (
+	// mininum max open files value
+	minRLimit = 1024
 )
 
+func GetSystemRLimit() (uint64, error) {
```
kennytm, line 33: add this function to `local_windows.go` too.

and maybe move `VerifyRLimit` out of `local_unix.go` back into `local.go`.

kennytm (commented):
rest LGTM

lightning/backend/local_unix.go
glorv: done! `VerifyRLimit` is always called, so the current implementation is ok. Add I added a check if check-requirement is disabled, also skip check rlimit, to match the logic in local_windows.

kennytm: I mean since VerifyRLimit calls GetSystemRLimit, the former is no longer directly system-dependent. But this is fine too.

kennytm (commented):
LGTM

ti-srebot (approved):

PR #505, opened 2020-12-02
backend: support stored generated columns in local/importer backends

kennytm:

Properly support stored generated columns in local/importer backends.

Unlike virtual generated columns, the value of stored generated columns are saved into the KV entries. This is computed before `AddRecord()`. Therefore, before this PR, all stored generated columns imported using local or importer backends become NULL.

We now fill in the stored generated columns by evaluating their expressions. 

### Check List 

Tests 

 - Integration test

Side effects

 - Increased code complexity

Related changes

# Release notes

Fixed: In local and importer backends, stored generated columns are properly filled in instead of being NULL.

### DIFF
--- a/lightning/backend/session.go
+++ b/lightning/backend/session.go
@@ -197,6 +197,7 @@ func newSession(options *SessionOptions) *session {
 	vars.StmtCtx.TimeZone = vars.Location()
 	vars.SetSystemVar("timestamp", strconv.FormatInt(options.Timestamp, 10))
 	vars.SetSystemVar(variable.TiDBRowFormatVersion, options.RowFormatVersion)
+	vars.SetSystemVar(variable.MaxAllowedPacket, strconv.FormatUint(variable.MaxOfMaxAllowedPacket, 10))
 	vars.TxnCtx = nil
 
 	s := &session{
--- a/lightning/backend/sql2kv.go
+++ b/lightning/backend/sql2kv.go
@@ -15,15 +15,19 @@ package backend
 
 import (
 	"math/rand"
+	"sort"
 
 	"github.com/pingcap/errors"
 	"github.com/pingcap/parser/model"
 	"github.com/pingcap/parser/mysql"
+	"github.com/pingcap/tidb/expression"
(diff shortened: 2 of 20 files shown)


lightning/backend/session.go
```diff
@@ -197,6 +197,7 @@ func newSession(options *SessionOptions) *session {
 	vars.StmtCtx.TimeZone = vars.Location()
 	vars.SetSystemVar("timestamp", strconv.FormatInt(options.Timestamp, 10))
 	vars.SetSystemVar(variable.TiDBRowFormatVersion, options.RowFormatVersion)
+	vars.SetSystemVar(variable.MaxAllowedPacket, strconv.FormatUint(variable.MaxOfMaxAllowedPacket, 10))
```
kennytm, line 200: (the `RPAD()` function relies on `@@max_allowed_packet`)

glorv:
LGTM

lightning/backend/sql2kv.go
```diff
@@ -246,6 +348,24 @@ func (kvcodec *tableKVEncoder) Encode(
...
+			evaluated, err := gc.expr.Eval(mutRow.ToRow())
+			if err != nil {
+				return nil, logEvalGenExprFailed(logger, row, col, err)
+			}
+			value, err := table.CastValue(kvcodec.se, evaluated, col, false, false)
+			if err != nil {
+				return nil, logEvalGenExprFailed(logger, row, col, err)
+			}
+			mutRow.SetDatum(gc.index, value)
+			record[gc.index] = value
```
lance6716, line 365: LGTM

Why need these two setter both🤔

lance6716 (approved):

lightning/backend/sql2kv.go
kennytm: @lance6716 Setting the MutRow is needed for cascading generated columns (think `c as b+1` & `a as b+1`, when calculating `c` we need to calculate `b` first). 
Setting the `record` is for inserting the result into the table. 
`MutRow` and `[]types.Datum` can't be cheaply converted between each other so we keep both of them.

PR #508, opened 2020-12-04
backend,restore: duplicate more important system variables from downstream

### LINKED ISSUE #504: Recognize `block_encryption_mode` from downstream
Lightning as of 4.0.9 always use the default `block_encryption_mode`. Since `block_encryption_mode` affects result of `aes_encrypt()` and `aes_decrypt()`, using the wrong mode may produce incorrect result for stored generated columns.

We should always copy the `block_encryption_mode` from downstream, just like `tidb_row_format_version`.

kennytm:

Fix #504.

Copy all sys vars that affects expression evaluation results in additional to `tidb_row_format_version`.

### Check List 

Tests 

 - Unit test
 - Integration test (will add integration test after #505 is merged)

Side effects

Related changes

### DIFF
--- a/lightning/glue/glue.go
+++ b/lightning/glue/glue.go
@@ -44,6 +44,7 @@ type SQLExecutor interface {
 	// same underlying connection
 	ExecuteWithLog(ctx context.Context, query string, purpose string, logger log.Logger) error
 	ObtainStringWithLog(ctx context.Context, query string, purpose string, logger log.Logger) (string, error)
+	QueryStringsWithLog(ctx context.Context, query string, purpose string, logger log.Logger) ([][]string, error)
 	Close()
 }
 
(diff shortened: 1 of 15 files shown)


kennytm:
the test failure in tidb backend is pingcap/tidb#21510 😓

lightning/glue/glue.go
```diff
@@ -44,6 +44,7 @@ type SQLExecutor interface {
 	// same underlying connection
 	ExecuteWithLog(ctx context.Context, query string, purpose string, logger log.Logger) error
 	ObtainStringWithLog(ctx context.Context, query string, purpose string, logger log.Logger) (string, error)
+	QueryStringsWithLog(ctx context.Context, query string, purpose string, logger log.Logger) ([][]string, error)
```
glorv, line 47: I'm a little concerned that we will add more QueryTypeWithLog functions here in the future.

kennytm: those other types can still be converted to string 🙃.

we can make it return `[][]interface{}` but so far such generalization isn't needed yet

glorv:
LGTM

overvenus:
LGTM

kennytm:
(manual approval needed)

overvenus (approved):

PR #358, opened 2020-07-24
restore: decouple restore/import/post-process 

glorv:
### What problem does this PR solve? 
Use channals to decouple restore/import/post-process phases to make each stage run parallel better.

### Check List 

Tests 

 - Unit test
 - Integration test
 - Manual test (add detailed scripts or steps below)
 - No code

Side effects

Related changes

PR #509, opened 2020-12-07
post-process: support run table analyze after all tables are finished 

glorv:
Add a new config `analyze-at-last` to allow run all table analyze tasks after all tables are finished. This is useful if TiKV instance's cpu is not rich (e.g. in standard dbaas env, TiKV is deployed with 8core cpu), thus the analyze operation will have significant impacts on SST ingest opeartions.

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Side effects

Related changes


kennytm:
how about #358 🤔

glorv:
> how about #358 🤔

Even based on #358, we still need to do extra work to allow delay analyze phase until all other tasks are finished. Since in our current design, post-process are always executed together.

kennytm:
Okay.

lightning/restore/restore.go
```diff
@@ -1230,17 +1270,20 @@ func (t *TableRestore) postProcess(ctx context.Context, rc *RestoreController, c
...
 			if err != nil {
-				return errors.Trace(err)
+				return false, errors.Trace(err)
 			}
 		}
+		cp.Status = CheckpointStatusChecksummed
 	}
 
 	// 5. do table analyze
+	doAnalyze := true
```
lance6716, line 1280: I'm a little confused about `doAnalyze` means this function do or caller should do, hope another name

lightning/restore/restore.go
```diff
@@ -814,7 +825,31 @@ func (rc *RestoreController) restoreTables(ctx context.Context) error {
...
+	default:
+	}
+
+	close(postProcessTaskChan)
+	// otherwise, we should run all tasks in the post-process task chan
+	for i := 0; i < rc.cfg.App.TableConcurrency; i++ {
+		wg.Add(1)
+		go func() {
+			for task := range postProcessTaskChan {
+				_, err := task.tr.postProcess(ctx, rc, task.cp, true)
```
lance6716, line 846: I think the last parameter means "can't delay".  if it's "can delay", I expect that this function may or may not run something in this turn.

glorv: replaced with a clearer name

Little-Wallace (approved):
LGTM

lance6716 (approved):
LGTM

PR #510, opened 2020-12-07
test: fix invalid failpoint and integration test

glorv:
Fix the wrong declaration of fail-point in integration test `local_backend` and update the definition of related fail-point

### Check List 

Tests 

 - Integration test

Side effects

Related changes


lightning/backend/local.go
```diff
@@ -949,23 +950,49 @@ loopWrite:
...
-							CurrentRegions: []*metapb.Region{region.Region}}
+						if retryCnt == 0 {
+							resp = &sst.IngestResponse{
+								Error: &errorpb.Error{
+									EpochNotMatch: &errorpb.EpochNotMatch{
+										CurrentRegions: []*metapb.Region{region.Region},
+									},
+								},
+							}
+							retryCnt++
```
3pointer, line 978: miss `errCnt ++` ?

glorv: Not really. Since `epochnotmatch` error will trigger a retry of write in the other loop, increase the local variable errCnt in the inner loop makes no sense.

3pointer: ok, so `retryCnt` is more like a bool variable.

lightning/backend/local.go
```diff
@@ -949,23 +950,49 @@ loopWrite:
...
+						}
+					}
+					if resp != nil {
+						err = nil
+						doIngest = false
 					}
 				})
+				if doIngest {
+					resp, err = local.Ingest(ctx, meta, region)
```
3pointer, line 987: How about inject failpoint after this line, and remove `doIngest`?

glorv: I don't understand? the logic in this pr is to avoid actually doing Ingest since the ingest action will likely succeed.

lightning/backend/local.go
glorv: yes, in the current logic we only let the error fail once. Maybe in the future, we will add more tests that need to fail more times.

lightning/backend/local.go
glorv: For some retryable error, we can directly retry the error, but a successful ingest will remove the source sst, thus the retry ingest will failed by sst file not found.

lance6716:
rest LGTM

could we use `1*return("epochnotmatch")` to avoid using `retryCnt` (also weaken the injection to globally once)

lance6716:
need address comment @glorv

glorv:
@lance6716  PTAL again

lance6716:
LGTM

kennytm (commented):
LGTM

ti-srebot (approved):

PR #516, opened 2020-12-09
backend/local: remove useless and buggy truncate key

glorv:
Remove the useless and buggy `truncateRowKey` call when generates region split keys because this action is not compatible with clustered index. And it's also useless even with clustered index disabled.

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Side effects

Related changes


kennytm (commented):
LGTM

lance6716 (approved):

ti-srebot (approved):

PR #119, opened 2019-01-16
Limit closed engine count for coordinating `write` and `import`

lonng:
### What problem does this PR solve? 

Limit closed engine count for coordinating `write` and `import`

### What is changed and how it works?

Apply a worker before call `engine.Close()` and release worker after `engine.Cleanup`

### Check List 

Tests 

 - Manual test (add detailed scripts or steps below)

Code changes

Side effects

Related changes

PR #520, opened 2020-12-10
Sort index rather than insert it into skiplist

Little-Wallace:
Improve the speed of writing index to pebble.

This PR allows IndexEngine store more data in memory, and then flush it when it exceed the size limit of memtable.
Pebble will sort the WriteBatch and flush it directly if the size of WriteBatch exceed the half of memtable.

### Check List 

Tests 

 - Unit test
 - Integration test

Related changes


lightning/backend/local.go
```diff
@@ -1193,50 +1196,35 @@ func (local *local) FetchRemoteTableModels(ctx context.Context, schemaName strin
...
-	wo := &pebble.WriteOptions{Sync: false}
-
-	size := int64(0)
-	for _, pair := range kvs {
-		wb.Set(pair.Key, pair.Val, wo)
-		size += int64(len(pair.Key) + len(pair.Val))
-	}
-	if err := wb.Commit(wo); err != nil {
-		return err
+	tmpPath := filepath.Join(local.localStoreDir, engineUUID.String())
```
glorv, line 1213: This dir is already used in L#417 for open pebble db, I think better change to another dir. And btw, since all `LocalWriter`s share the same dir, just init it in the `OpenEngine` function is better

lightning/backend/backend.go
```diff
@@ -265,25 +260,42 @@ func (engine *OpenedEngine) Flush() error {
 
 // WriteRows writes a collection of encoded rows into the engine.
 func (engine *OpenedEngine) WriteRows(ctx context.Context, columnNames []string, rows Rows) error {
-	var err error
+	if rows.Len() == 0 {
```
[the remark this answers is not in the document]
Little-Wallace: This is only for a test....

lightning/backend/local.go
Little-Wallace: OK

lightning/backend/local.go
```diff
@@ -1510,3 +1498,181 @@ func (s *sizeProperties) iter(f func(p *rangeProperty) bool) {
...
+	produceWg         sync.WaitGroup
+	consumeWg         sync.WaitGroup
+	kvsChan           chan []common.KvPair
+	sstDir            string
+	memtableSizeLimit int64
+}
+
+func (w *LocalWriter) isSorted(kvs []common.KvPair) bool {
+	if len(kvs) <= 1 {
+		return false
```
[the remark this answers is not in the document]
Little-Wallace: Because if it is only one, maybe the keys come soon will not be in order.

lightning/backend/local.go
```diff
@@ -1510,3 +1498,181 @@ func (s *sizeProperties) iter(f func(p *rangeProperty) bool) {
...
+	totalSize := int64(0)
+	var writer *sstable.Writer = nil
+	//var wb *pebble.Batch = nil
+	var wb []common.KvPair
+	var filePath string
+	defer w.consumeWg.Done()
+	var err error
+	for kvs := range w.kvsChan {
+		atomic.AddInt64(&w.local.Length, int64(len(kvs)))
+		if len(wb) == 0 && w.isSorted(kvs) {
```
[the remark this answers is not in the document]
Little-Wallace: No.... We can save more memory by flush it into sst at once.

lightning/backend/local.go
```diff
@@ -1510,3 +1498,181 @@ func (s *sizeProperties) iter(f func(p *rangeProperty) bool) {
...
+				wb = kvs
+			} else {
+				wb = append(wb, kvs...)
+			}
+
+			if batchSize > w.memtableSizeLimit {
+				if err := w.flushKVs(wb); err != nil {
+					w.writeErr.Set(err)
+					return
+				}
```
gozssky, line 1590: Since the wb have been flushed to disk. Does the wb need to be reset? Something like `wb = wb[:0]`.

lightning/backend/local.go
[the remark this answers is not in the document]
glorv: There are two scenarios that can lead to only one key in a batch:
- The kv is quite big, exceed the batch send threshold(default to 64k)
- it is the last key in the chunk

In either case, I think it's ok to still do the sorted check to avoid adding one-key wb.

Or， maybe we should delay the sst writers phase until the sorted kv paired are large enough to be flushed, thus we can always do check for all keys

lightning/backend/local.go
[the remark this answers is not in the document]
glorv: But in the unordered scenario, we will still consume a larger amount of memory, so I think we can do the same thing with ordered keys if no performance loss.

lightning/backend/local.go
Little-Wallace: Yes... It is a bug.

lightning/backend/local.go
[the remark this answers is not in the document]
Little-Wallace: Obiviously, sort a large of keys will cost two much time.....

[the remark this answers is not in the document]
Little-Wallace: And if they are unordered, we will find it after only compare a few keys in front.

lightning/backend/local.go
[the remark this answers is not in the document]
Little-Wallace: Ok, I will check it.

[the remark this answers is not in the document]
Little-Wallace: The region concurrency will be large. It is not a good idea to keep too much keys in memory...

lightning/backend/local.go
```diff
@@ -1510,3 +1492,180 @@ func (s *sizeProperties) iter(f func(p *rangeProperty) bool) {
...
+		return err
+	}
+	sort.Slice(kvs, func(i, j int) bool {
+		return bytes.Compare(kvs[i].Key, kvs[j].Key) < 0
+	})
+	if _, err := writeKVs(writer, kvs); err != nil {
+		return err
+	}
+	if err := writer.Close(); err != nil {
+		return nil
```
glorv, line 1635: why ignores error here?

lightning/backend/local.go
```diff
@@ -77,6 +79,7 @@ const (
 
 	defaultPropSizeIndexDistance = 4 * 1024 * 1024 // 4MB
 	defaultPropKeysIndexDistance = 40 * 1024
+	IndexEngineID                = -1
```
glorv, line 82: this const is not used in package `backend`, so there is no need to move it here

lightning/backend/backend.go
```diff
@@ -388,3 +397,16 @@ type Rows interface {
 	// capacity with the current instance. The typical usage is `x = x.Clear()`.
 	Clear() Rows
 }
+
+type EngineWriter interface {
```
glorv, line 401: How `Engine` communicates with `EngineWriter`. In #493, the disk quota checking routine may trigger force flush and import action if the disk is nearly full, so each engine write should support force flush then

lightning/restore/restore.go
```diff
@@ -1116,19 +1127,16 @@ func (t *TableRestore) restoreEngine(
...
-	dataWorker := rc.closedEngineLimit.Apply()
 	closedDataEngine, err := dataEngine.Close(ctx)
-	// For local backend, if checkpoint is enabled, we must flush index engine to avoid data loss.
-	// this flush action impact up to 10% of the performance, so we only do it if necessary.
 	if err == nil && rc.cfg.Checkpoint.Enable && rc.isLocalBackend() {
 		if err = indexEngine.Flush(); err != nil {
 			// If any error occurred, recycle worker immediately
-			rc.closedEngineLimit.Recycle(dataWorker)
-			return nil, nil, errors.Trace(err)
+			return nil, errors.Trace(err)
```
glorv, line 1137: This `Flush` seems can not promise all in pending kvs are flush into pebble now

Little-Wallace: No. I use sync.WaitGroup to make sure that all writer has finished their works.

lightning/backend/local.go
Little-Wallace: Yes....I changed it for other reason and it does need now. I will revert it.

lightning/backend/local.go
Little-Wallace: I have fixed it.

glorv:
@kennytm PTAL

glorv:
LGTM

lightning/backend/backend.go
```diff
@@ -265,25 +260,39 @@ func (engine *OpenedEngine) Flush() error {
...
-	var err error
+	writer, err := engine.backend.LocalWriter(ctx, engine.uuid)
+	if err != nil {
+		return err
+	}
+	if err = writer.AppendRows(ctx, engine.tableName, columnNames, engine.ts, rows); err != nil {
+		return err
+	}
+	return writer.Close()
+}
```
kennytm, line 271: maybe just remove the whole method?

lightning/backend/backend.go
```diff
@@ -388,3 +397,16 @@ type Rows interface {
...
+	AppendRows(
+		ctx context.Context,
+		tableName string,
+		columnNames []string,
+		commitTS uint64,
+		rows Rows,
+	) error
+	Close() error
+	AddProducer()
+	Done()
```
kennytm, line 411: i think the `EngineWriter` interface doesn't need `AddProducer` and `Done` methods.
the `produceWg WaitGroup` could just be a member of `LocalEngineWriter`.

lightning/restore/restore.go
```diff
@@ -1057,23 +1055,26 @@ func (t *TableRestore) restoreEngine(
...
 	engineID int32,
 	cp *EngineCheckpoint,
-) (*kv.ClosedEngine, *worker.Worker, error) {
+) (*kv.ClosedEngine, error) {
 	if cp.Status >= CheckpointStatusClosed {
-		w := rc.closedEngineLimit.Apply()
 		closedEngine, err := rc.backend.UnsafeCloseEngine(ctx, t.tableName, engineID)
 		// If any error occurred, recycle worker immediately
 		if err != nil {
-			rc.closedEngineLimit.Recycle(w)
```
kennytm, line 1066: the `closedEngineLimit` was introduced in #119 to put a "back-pressure" if Importer is slower than Lightning. i'm not sure if we should remove the worker entirely.

> closed-engine: 一个 Lightning batch 都对应了一个 Importer 的 Engine，如果一个 batch 完全写入到 Importer，就会关闭这个 Engine，直到这个 Engine Import 到 TiKV 之后才会对这个 Engine 进行清理，如果有 closed engine 的 Idle 数量少，表示 Lightning 的速度比 Importer 的速度快，等 closed engine 的 Idle 数量完全为 0 时，Lightning 会暂停写入。

lightning/backend/backend.go
Little-Wallace: There is still some tests using this method.

lightning/backend/backend.go
Little-Wallace: Ok. I will refactor it.

lightning/restore/restore.go
Little-Wallace: Thanks. I think we can import KV async and do not block this goroutine so that it can import other table. I have talked this with @glorv . But I will revert it in this PR in order not to introduce irrelevant changes.

glorv: From the current implementation, `closedEngineLimit` mainly affect on how many engines than can be imported in parallel. This pr just move the apply action from close engine to before start import. Does this really matter 🤔？ @kennytm

lightning/backend/backend.go
Little-Wallace: Done

lightning/backend/backend.go
kennytm: if they are only used in tests, they can be converted to use the local writer.

(since i'm not so fond of apparently leaking the LocalWriter (not calling Close) when AppendRows returns an error, but it isn't worth "fixing" if no one actually uses it)

lightning/restore/restore.go
kennytm: if they have the same effect then LGTM on this part.

lightning/backend/backend.go
Little-Wallace: Yes, I will fix it and add Close.

kennytm:
LGTM

ti-srebot (approved):

PR #523, opened 2020-12-14
backend/local: fix next key

glorv:
in tikv release-4.0, tikv will force truncate all row keys, so our currently truncate key is not valid according to this logic. 
See: https://github.com/tikv/tikv/blob/f7f22f70e1585d7ca38a59ea30e774949160c3e8/components/raftstore/src/coprocessor/split_observer.rs#L36-L41

If the key is a row key, use handle.Next() to generate the next key, else, append a 0x00 to the key otherwise.

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Side effects

Related changes

### DIFF
--- a/lightning/backend/local.go
+++ b/lightning/backend/local.go
@@ -41,8 +41,8 @@ import (
 	"github.com/pingcap/kvproto/pkg/kvrpcpb"
 	"github.com/pingcap/kvproto/pkg/metapb"
 	"github.com/pingcap/parser/model"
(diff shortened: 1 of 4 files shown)


3pointer (commented):
LGTM

kennytm (commented):
LGTM

ti-srebot (approved):

glorv:
/run-all-tests tikv=release-4.0 pd=release-4.0 tidb=release-4.0 importer=release-4.0

PR #524, opened 2020-12-14
backend/local: skip split regions if engine total size is smaller than region size

glorv:
Skip split regions there is only one range within the engine.

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Side effects

Related changes


lightning/backend/local.go
```diff
@@ -1122,19 +1122,27 @@ func (local *local) ImportEngine(ctx context.Context, engineUUID uuid.UUID) erro
...
+					log.ShortError(err), zap.Int("retry", i))
+			}
+			if err != nil {
+				log.L().Error("split & scatter ranges failed", zap.Stringer("uuid", engineUUID), log.ShortError(err))
+				return err
+			}
+		} else {
+			log.L().Info("engine total size is smaller than region split size, skip split regions",
+				zap.Stringer("uuid", engineUUID), zap.Int64("size", lf.TotalSize),
+				zap.Int64("count", lf.Length))
```
kennytm, line 1145: should we still scatter 🤔

glorv: I think not since we don't create any new region, pd should be responsible for the distribution of the existed regions.

kennytm: if we create lots of small tables on a fresh cluster will this cause all data still places on the same 3 stores? or will CREATE TABLE scatter?

Little-Wallace: Yes. Maybe we shall still scatter it.

glorv: Currently `preSplitAndScatter` depend on the session variable `tidb_scatter_region` (default is 'OFF') for whether scatter or not. So maybe we can set this variable to 'ON' at restoreSchema. 
See: https://github.com/pingcap/tidb/blob/46fa1628896fce4e30d990b9508479bc7e3add2b/ddl/ddl_api.go#L1869-L1893

kennytm: @glorv SGTM.

glorv: There is still a problem, the `tidb_scatter_region` is a global variable, so if we switch it to off before restoring schemas and on after finished, how to ensure it is restored if lightning exit before finishing restoreSchema. 

I added a feature request to tidb. https://github.com/pingcap/tidb/issues/21991

kennytm: @glorv what if we add `PRE_SPLIT_REGIONS=1` to the `CREATE TABLE` statement

kennytm (commented):
LGTM

lightning/backend/local.go
```diff
@@ -1181,10 +1181,14 @@ func (local *local) ImportEngine(ctx context.Context, engineUUID uuid.UUID) erro
 		log.L().Info("start import engine", zap.Stringer("uuid", engineUUID),
 			zap.Int("ranges", len(ranges)))
 
+		// if all the kv can fit in one region, skip split regions. TiDB will split one region for
+		// the table when table is created.
+		needSplit := len(ranges) > 0 || lf.TotalSize > local.regionSplitSize || lf.Length > regionMaxKeyCount
```
lance6716, line 1186: is `needSplit := len(ranges) > 1` enough? seems `readAndSplitIntoRange` has checked `regionSplitSize` and `regionMaxKeyCount`

https://github.com/pingcap/tidb-lightning/blob/d9254b1deea8607f6ffd0f94a9cae3135d272745/lightning/backend/local.go#L811-L815

and `splitRangeBySizeProps` uses a tighter `keysLimit` so it will return more than 1 `ranges`

glorv: splitRangeBySizeProps may still generate ranges that exceed regionSplitSize, so if the split is triggered by remain ranges, it may be a single range but the range size is exceeded.

lance6716: I see. Now I suggest

```suggestion
		needSplit := len(ranges) > 1 || lf.TotalSize > local.regionSplitSize || lf.Length > regionMaxKeyCount
```

lance6716 (commented):
rest LGTM

lightning/backend/local.go
glorv: done

lance6716:
LGTM

ti-srebot (approved):

PR #530, opened 2020-12-22
restore: fix error lost in create schema

3pointer:
Fix https://github.com/pingcap/tidb/issues/21827

break create loop immediately when error encountered

### Check List 

Tests 

Related changes

Release note:

- Fix the issue that lost create table error during multiple tables created.

### DIFF
--- a/lightning/restore/tidb.go
+++ b/lightning/restore/tidb.go
@@ -17,7 +17,6 @@ import (
 	"context"
 	"database/sql"
 	"fmt"
-	"sort"
 	"strconv"
 	"strings"
 
@@ -30,6 +29,7 @@ import (
 	"github.com/pingcap/parser/model"
 	"github.com/pingcap/parser/mysql"
 	"github.com/pingcap/parser/terror"
+
 	"github.com/pingcap/tidb-lightning/lightning/glue"
 
 	. "github.com/pingcap/tidb-lightning/lightning/checkpoints"
@@ -156,15 +156,8 @@ func InitSchema(ctx context.Context, g glue.Glue, database string, tablesSchema 
 
 	task := logger.Begin(zap.InfoLevel, "create tables")
 	var sqlCreateStmts []string
-	tables := make([]string, 0, len(tablesSchema))
-	for tbl := range tablesSchema {
-		tables = append(tables, tbl)
-	}
-	// sort tables for failpoint test
-	sort.Strings(tables)
 loopCreate:
-	for i, tbl := range tables {
-		sqlCreateTable := tablesSchema[tbl]
+	for tbl, sqlCreateTable := range tablesSchema {
 		task.Debug("create table", zap.String("schema", sqlCreateTable))
 
 		sqlCreateStmts, err = createTableIfNotExistsStmt(g.GetParser(), sqlCreateTable, database, tbl)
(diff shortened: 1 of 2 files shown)


lightning/restore/tidb.go
```diff
@@ -154,7 +156,15 @@ func InitSchema(ctx context.Context, g glue.Glue, database string, tablesSchema
 
 	task := logger.Begin(zap.InfoLevel, "create tables")
 	var sqlCreateStmts []string
-	for tbl, sqlCreateTable := range tablesSchema {
+	tables := make([]string, 0, len(tablesSchema))
+	for tbl := range tablesSchema {
+		tables = append(tables, tbl)
+	}
+	// sort tables for failpoint test
```
glorv, line 163: I think a better solution is to use regexp to match the execute result in failpoint test

lightning/restore/tidb.go
```diff
@@ -168,10 +171,17 @@ func InitSchema(ctx context.Context, g glue.Glue, database string, tablesSchema
 				ctx,
 				s,
 				"create table",
-				logger.With(zap.String("table", common.UniqueTable(database, tbl))),
+				logger.With(zap.String("table", common.UniqueTable(database, tblSchema.tblName))),
 			)
+			failpoint.Inject("sqlCreateStmts", func() {
+				if i == 0 {
```
glorv, line 177: if the count matters, I think you can add an int parameter to this fail-point to determine at which condition, it should return an error

lightning/restore/restore.go
```diff
@@ -317,12 +322,14 @@ func (rc *RestoreController) restoreSchema(ctx context.Context) error {
 
 		for _, dbMeta := range rc.dbMetas {
 			task := log.With(zap.String("db", dbMeta.Name)).Begin(zap.InfoLevel, "restore table schema")
-
-			tablesSchema := make(map[string]string)
+			tablesSchemas := make([]tableSchema, 0, len(dbMeta.Tables))
```
glorv, line 325: Please add a comment here why we don't use a map to hold these data, it can help others to understand it better

3pointer: Done, PTAL

lightning/restore/tidb.go
```diff
@@ -168,10 +171,17 @@ func InitSchema(ctx context.Context, g glue.Glue, database string, tablesSchema
...
+					err = errors.Errorf("create %s failed", tblSchema.tblName)
+				} else {
+					err = nil
```
kennytm, line 180: you shouldn't reset the `err` to nil here, as `ExecuteWithLog` may really fail

lightning/restore/restore.go
kennytm: how looping over a map makes injection harder? you could make the failpoint return a table name.

lightning/restore/tidb_test.go
```diff
@@ -224,6 +225,25 @@ func (s *tidbSuite) TestInitSchemaSyntaxError(c *C) {
...
 }
 
+func (s *tidbSuite) TestInitSchemaErrorLost(c *C) {
+	ctx := context.Background()
+
+	s.mockDB.
+		ExpectExec("CREATE DATABASE IF NOT EXISTS `db`").
+		WillReturnResult(sqlmock.NewResult(1, 1))
+	s.mockDB.
+		ExpectClose()
```
kennytm, line 235: actually you could do

```go
s.mockDB. 
    ExpectExec("create table.*").
    WillReturnError(errors.New(...))
```

to inject an error. no need to introduce a failpoint
```

kennytm (commented):
LGTM

glorv:
LGTM

ti-srebot (approved):

PR #531, opened 2020-12-23
backend/local: add encode for retry scan region keys

glorv:
Fix the bug that the retry split key is not encoded, which causes scanned regions are incorrect.

- Fix the retry split scan keys bug 
- Add more error checks and logs
- Add some unit tests

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Side effects

Related changes

Release Note
- Fix the bug that the retry scan region key is not encoded, which causes the retryed region is not split.

### DIFF
--- a/lightning/backend/localhelper.go
+++ b/lightning/backend/localhelper.go
@@ -42,6 +42,9 @@ const (
 var (
 	// the max keys count in a batch to split one region
 	maxBatchSplitKeys = 4096
+	// the base exponential backoff time
+	// the variable is only changed in unit test for running test faster.
+	splitRegionBaseBackOffTime = time.Second
 )
 
 // TODO remove this file and use br internal functions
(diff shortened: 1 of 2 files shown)


lightning/backend/localhelper.go
```diff
@@ -85,13 +87,31 @@ func (local *local) SplitAndScatterRegionByRanges(ctx context.Context, ranges []
...
 		regionMap := make(map[uint64]*split.RegionInfo)
 		for _, region := range regions {
 			regionMap[region.Region.GetId()] = region
 		}
 
 		var splitKeyMap map[uint64][][]byte
 		if len(retryKeys) > 0 {
+			firstKeyEnc := codec.EncodeBytes([]byte{}, retryKeys[0])
+			lastKeyEnc := codec.EncodeBytes([]byte{}, retryKeys[len(retryKeys)-1])
+			if bytes.Compare(firstKeyEnc, regions[0].Region.StartKey) < 0 || !beforeEnd(lastKeyEnc, regions[len(regions)-1].Region.EndKey) {
```
Little-Wallace, line 108: Why compare firstKey and regions[0]?

glorv: fistKey is the smallest key, so it should be included in region[0]; and lastKey in region[-1]

Little-Wallace: It seems that region.StartKey without encode can not compare to other encoded key directly?

Little-Wallace: Just like this:https://github.com/pingcap/tidb-lightning/blob/master/lightning/backend/local.go#L932

Little-Wallace (approved):
LGTM

3pointer (approved):
LGTM

PR #537, opened 2020-12-25
restore: add error retry for checksum by tikv

glorv:
Add error retry for checksum with TiKV executor to be more robust with the timeout error.

- Add retry for checksum executor, with each retry, decrease the dist SQL scan concurrency by half.

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Side effects

Related changes


lightning/restore/checksum.go
```diff
@@ -267,18 +270,33 @@ func (e *tikvChecksumManager) checksumDB(ctx context.Context, tableInfo *TidbTab
...
+				Schema:     tableInfo.DB,
+				Table:      tableInfo.Name,
+				Checksum:   execRes.Checksum,
+				TotalBytes: execRes.TotalBytes,
+				TotalKVs:   execRes.TotalKvs,
+			}, nil
+		}
+
+		log.L().Warn("remote checksum failed", zap.String("db", tableInfo.DB),
+			zap.String("table", tableInfo.Name), zap.Error(err),
```
kennytm, line 292:
```suggestion
			zap.String("table", tableInfo.Name), log.ShortError(err),
```

lightning/restore/checksum.go
```diff
@@ -267,18 +270,33 @@ func (e *tikvChecksumManager) checksumDB(ctx context.Context, tableInfo *TidbTab
 		return nil, errors.Trace(err)
 	}
 
-	res, err := executor.Execute(ctx, e.client, func() {})
-	if err != nil {
-		return nil, errors.Trace(err)
+	distSQLScanConcurrency := int(e.distSQLScanConcurrency)
+	for i := 0; i < 3; i++ {
```
kennytm, line 274: turn the "3" into a constant

kennytm (commented):
rest LGTM

kennytm:
lgtm

lightning/restore/checksum.go
```diff
@@ -10,14 +10,15 @@ import (
 	"time"
 
 	"github.com/google/uuid"
-
 	"github.com/pingcap/br/pkg/checksum"
+	"github.com/pingcap/br/pkg/utils"
```
lance6716, line 14: we might use `github.com/cznic/mathutil`, because I remember this path will cause a cycle import after lightning via SQL.

lightning/restore/checksum.go
```diff
@@ -267,18 +272,33 @@ func (e *tikvChecksumManager) checksumDB(ctx context.Context, tableInfo *TidbTab
...
+			return &RemoteChecksum{
+				Schema:     tableInfo.DB,
+				Table:      tableInfo.Name,
+				Checksum:   execRes.Checksum,
+				TotalBytes: execRes.TotalBytes,
+				TotalKVs:   execRes.TotalKvs,
+			}, nil
+		}
+
+		log.L().Warn("remote checksum failed", zap.String("db", tableInfo.DB),
```
lance6716, line 293: could we distinguish timeout error and only retry for that?

lance6716 (commented):
rest LGTM

lightning/restore/checksum.go
glorv: Add check whether is the error is retryable

lightning/restore/checksum.go
glorv: It's weird that this dep causes cyclic deps? Since this file allow depends on other br pkgs

ti-srebot (approved):

Issue #518, opened 2020-12-10
Lightning report `[tikv:9002]TiKV server timeout` when analyze enabled


shuijing198799:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do? If possible, provide a recipe for reproducing the error.
Lightning will failed for `[tikv:9002]TiKV server timeout` when analyze enabled
2. What did you expect to see?
lightning import completed
3. What did you see instead?
failed
4. Versions of the cluster

    - TiDB-Lightning version (run `tidb-lightning -V`):

        ```
Release Version: v4.0.8-42-g59cc48d
Git Commit Hash: 59cc48d738c237c1bedfa2c758f905828369736e
Git Branch: master
UTC Build Time: 2020-11-26 12:45:03
Go Version: go version go1.13.8 linux/amd64
        ```
        
    - TiKV version (run `tikv-server -V`):

        ```
Release Version:   4.0.8
Edition:           Community
Git Commit Hash:   30001da180ba2408b905df061a10a73e482a8172
Git Commit Branch: HEAD
UTC Build Time:    2020-12-07 09:03:56
Rust Version:      rustc 1.42.0-nightly (0de96d37f 2019-12-19)
Enable Features:   jemalloc mem-profiling portable sse protobuf-codec
Profile:           release
        ```

cyliu0:
Reproduced on 4.0.9

cyliu0:
Reproduced on 4.0.9 while importing 4TB tpcc data

Little-Wallace:
Closed by https://github.com/pingcap/tidb-lightning/pull/540

PR #540, opened 2020-12-28
post-process: allow run checksum at last and restrict the number of checksum jobs

glorv:
- Also allow run checksum jobs at last
- Restrict the number of post-process jobs by config field `checksum-table-concurrency` and set its default value to 4 since 16 is a very big value.
- Change default config `distsql-scan-concurrency` from 100 to 15

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Side effects
 - change Some default config, this may cause checksum and analyze to run slower in cluster with abandunt CPU resources.
   - replace config field `analyze-at-last` with `checksum-at-last`

Related changes


lightning/config/config.go
```diff
@@ -337,7 +337,7 @@ func NewConfig() *Config {
 			BuildStatsConcurrency:      20,
 			DistSQLScanConcurrency:     100,
```
Little-Wallace, line 338: `DistSQLScanConcurrency ` will decide the real concurrency  how many requests to each region are executed concurrently.

Little-Wallace: It may not exceed the total thread number, which is sum of the size of coprocessor pool of each TiKV instance of the whole Cluster,

lightning/config/config.go
```diff
@@ -204,11 +209,11 @@ func (t PostOpLevel) String() string {
...
-	Level1Compact bool        `toml:"level-1-compact" json:"level-1-compact"`
-	Compact       bool        `toml:"compact" json:"compact"`
-	Checksum      PostOpLevel `toml:"checksum" json:"checksum"`
-	Analyze       PostOpLevel `toml:"analyze" json:"analyze"`
-	AnalyzeAtLast bool        `toml:"analyze-at-last" json:"analyze-at-last"`
+	Level1Compact  bool        `toml:"level-1-compact" json:"level-1-compact"`
+	Compact        bool        `toml:"compact" json:"compact"`
+	Checksum       PostOpLevel `toml:"checksum" json:"checksum"`
+	Analyze        PostOpLevel `toml:"analyze" json:"analyze"`
+	ChecksumAtLast bool        `toml:"checksum-at-last" json:"checksum-at-last"`
```
Little-Wallace, line 216: Do we need to set initial value `true` in `NewConfig`?

tidb-lightning.toml
```diff
@@ -247,8 +248,8 @@ level-1-compact = false
 # if set true, compact will do full compaction to tikv data.
 # if this setting is missing, the default value is false.
 compact = false
-# if set to true, lightning will analyze all table together at last
-analyze-at-last = false
+# if set to true, lightning will run checksum and analyze for all tables together at last
+checksum-at-last = false
```
kennytm, line 252: renaming this option is a breaking change. has #509 been included in any release?

glorv: No, the newest v4.0.9/v3.0.20 doesn't contain this commit

kennytm: in that case maybe choosing a word which represents both "checksum" and "analyze" is better (e.g. `post-process-at-last`). or maybe declare "analyze" means "checksum" as well 🙃

Little-Wallace (approved):
LGTM

tidb-lightning.toml
Little-Wallace: I agree with `post-process-at-last`. I think it is not a configuration that could be changed by customer. It could only be adjust by DBA or develop.

zhangjinpeng1987:
Any related issues for this PR?

Little-Wallace:
Related to https://github.com/pingcap/tidb-lightning/issues/518
root case (maybe one part of it): Lightning will sent many coprocessor requests to TiKV (one for each region) for `checksum` and `analyze`, the number of which is equal to `distsql-scan-concurrency`. If TiKV can not consume these requets as soon as possible, the latency of some requests may exceed 10min, which is the timeout duration of gRPC client in Lightning, so lightning will get an error of `TiKV server timeout`.

kennytm (dismissed):
LGTM

lightning/restore/restore.go
```diff
@@ -1236,9 +1238,14 @@ func (t *TableRestore) importEngine(
 
 // postProcess execute rebase-auto-id/checksum/analyze according to the task config.
 //
-// if the parameter forceAnalyze to true, postProcess force run analyze even if the analyze-at-last config is true.
-// And if analyze phase is skipped, the first return value will be true.
-func (t *TableRestore) postProcess(ctx context.Context, rc *RestoreController, cp *TableCheckpoint, forceAnalyze bool) (bool, error) {
+// if the parameter forcePostProcess to true, postProcess force run checksum and analyze even if the
+// analyze-at-last config is true. And if this two phases are skipped, the first return value will be true.
```
lance6716, line 1242:
```suggestion
// post-process-at-last config is true. And if this two phases are skipped, the first return value will be true.
```

lightning/restore/restore.go
```diff
@@ -1272,44 +1279,54 @@ func (t *TableRestore) postProcess(ctx context.Context, rc *RestoreController, c
...
+				// 4. do table checksum
+				var localChecksum verify.KVChecksum
+				for _, engine := range cp.Engines {
+					for _, chunk := range engine.Chunks {
+						localChecksum.Add(&chunk.Checksum)
+					}
+				}
+				t.logger.Info("local checksum", zap.Object("checksum", &localChecksum))
+				err := t.compareChecksum(ctx, localChecksum)
+				// witch post restore level 'optional', we will skip checksum error
```
lance6716, line 1301:
```suggestion
				// with post restore level 'optional', we will skip checksum error
```

lance6716 (commented):
rest LGTM

lance6716:
LGTM

ti-srebot (approved):

PR #543, opened 2021-01-04
Compatible for disk quota

Little-Wallace:
Signed-off-by: Little-Wallace <bupt2013211450@gmail.com>

Compatible for https://github.com/pingcap/tidb-lightning/pull/493

### Check List 

Tests 

 - Integration test

Side effects

 - No

Related changes

- No


lance6716:
what problem does this PR solve 🤔

Little-Wallace:
> what problem does this PR solve 🤔

I just add some lock so that `diskquota` could change the db instance of `LocalFile` after `ImportAndResetEngine`.

Little-Wallace:
@lance6716 @kennytm Any comment?

lightning/backend/local.go
```diff
@@ -1613,13 +1680,20 @@ func (w *LocalWriter) writeRowsLoop() {
 			w.writeErr.Set(err)
 			return
 		}
-		if err := w.local.db.Ingest([]string{filePath}); err != nil {
-			w.writeErr.Set(err)
-			return
-		}
+		w.local.lockForImport(importMutexStateFlush)
+		err := w.local.db.Ingest([]string{filePath})
+		// The following two variable should be changed with holding mutex
```
lance6716, line 1685: Seems after we `ResetEngine`, we should also reopen the `LocalWriter` 😂 need to change a lot

lance6716 (commented):
is this PR caused by #493 is too long and want to fix something in #493? I'm not sure how we should cooperate in this situation (maybe create a new branch rather than merge into master?)

lightning/backend/local.go
Little-Wallace: No. We do not need to reopen `LocalWriter`.  Because LocalWriter holds `LocalFile` rather than `pebble.DB`. So we only need to update the field `LocalFile.db` with holding mutex.

lightning/backend/local.go
```diff
@@ -1613,13 +1680,20 @@ func (w *LocalWriter) writeRowsLoop() {
 			w.writeErr.Set(err)
 			return
 		}
-		if err := w.local.db.Ingest([]string{filePath}); err != nil {
-			w.writeErr.Set(err)
-			return
-		}
+		w.local.lockForImport(importMutexStateFlush)
```
glorv, line 1683: `lockForImport` will set isImporting to true, thus this will impact the force importing logic since force importing will skip engine in importing state. I think it's better to just lock but not set the atomic var

lightning/backend/local.go
```diff
@@ -1639,7 +1713,10 @@ func (w *LocalWriter) flushKVs() error {
 		return err
 	}
 	w.writeBatch = w.writeBatch[:0]
-	return w.local.db.Ingest([]string{filePath})
+	w.local.lockForImport(importMutexStateFlush)
```
glorv, line 1716: ditto

lightning/backend/local.go
```diff
@@ -447,7 +496,12 @@ func (local *local) OpenEngine(ctx context.Context, engineUUID uuid.UUID) error
 	if err != nil {
 		return err
 	}
-	local.engines.Store(engineUUID, &LocalFile{localFileMeta: meta, db: db, Uuid: engineUUID})
+	if e, ok := local.engines.Load(engineUUID); ok {
+		engine := e.(*LocalFile)
+		engine.db = db
```
glorv, line 501: should also reset meta here

lightning/backend/local.go
Little-Wallace: Why? I do not think `force importing` shall choose one db which is ingesting a file. This duration is short and if `force importing` want to choose this db, it shall wait it to finish ingesting.

lance6716: IMO importing atomic var is used to "take a look at it" to decide if it could be further force importing, when further force importing, lightning will do aquire the lock.

so setting importing atomic var in this write phase will cause a "false positive" (I'll check my logic later 😂 )

Little-Wallace: If it meet a `importMutexStateFlush`, it will skip forcing this engineFile importing. This is a very unlikely event. Even if it happens, we just left it force-import in next time.

Little-Wallace: I have change it into another state `NoLock`.

Little-Wallace:
@glorv @lance6716 PTAL

glorv:
@Little-Wallace  There is still one unresolved comment🤔

lance6716 (commented):
rest LGTM

lightning/backend/local.go
Little-Wallace: OK.

lightning/backend/local.go
```diff
@@ -154,6 +168,21 @@ func (e *LocalFile) getSizeProperties() (*sizeProperties, error) {
 	return sizeProps, nil
 }
 
+func (e *LocalFile) isLocked() bool {
+	return atomic.LoadInt32(&e.isImportingAtomic) != 0
```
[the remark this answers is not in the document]
Little-Wallace: OK

lightning/backend/local.go
```diff
@@ -1608,18 +1676,22 @@ func (w *LocalWriter) writeRowsLoop() {
...
-			w.writeErr.Set(err)
-			return
-		}
-		if err := w.local.db.Ingest([]string{filePath}); err != nil {
-			w.writeErr.Set(err)
-			return
+		err := writer.Close()
+		if err == nil {
+			err = w.local.db.Ingest([]string{filePath})
+			// The following two variable should be changed with holding mutex
```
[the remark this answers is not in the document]
Little-Wallace: OK, I will add a comment. Because another thread will reset the whole meta object rather than update this two atomic number.

lightning/backend/local.go
```diff
@@ -447,7 +497,12 @@ func (local *local) OpenEngine(ctx context.Context, engineUUID uuid.UUID) error
 	if err != nil {
 		return err
 	}
-	local.engines.Store(engineUUID, &LocalFile{localFileMeta: meta, db: db, Uuid: engineUUID})
+	if e, ok := local.engines.Load(engineUUID); ok {
+		engine := e.(*LocalFile)
+		engine.db = db
+	} else {
+		local.engines.Store(engineUUID, &LocalFile{localFileMeta: meta, db: db, Uuid: engineUUID, isImportingAtomic: 0})
```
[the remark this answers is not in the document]
Little-Wallace: But I can not pass CI because of uninitialize isImportingAtomic.....

lightning/backend/local.go
```diff
@@ -305,6 +334,26 @@ func NewLocalBackend(
 	return MakeBackend(local), nil
 }
 
+// lockForImport locks the local file for importing.
+func (local *local) lockForImport(engineId uuid.UUID, state importMutexState) bool {
+	if e, ok := local.engines.Load(engineId); ok {
+		engine := e.(*LocalFile)
+		engine.lockForImport(state)
```
kennytm, line 341: @lance6716 we don't need UUID-level lock?

Little-Wallace: #493 will reopen this engineFile. So we need to lock it when resetengine.

lance6716: Need. I just think this PR should be merged into master, and #493 will adjust its content later 😂

lance6716: i'm confused in https://github.com/pingcap/tidb-lightning/pull/543#pullrequestreview-561469974

kennytm: @Little-Wallace maybe you could explain a bit 😅

Little-Wallace: Emmmmm.  #493  will open a new db instance and update it to `localFile`. But this db instance may be called in another goroutine, which is `writeRowsLoop` and try to ingest a sst file into itself. So we must add a lock to prevent data race.

lance6716: if #493 will split `ResetEngine` and `CleanupEngine` to two behaviour, which the former didn't remove `LocalFile` from `local.engines`, I think we don't need UUID-level lock then 😂 because `LocalFile` is not changed after bounded to UUID in map

Little-Wallace: Oh. Yes. We only need to lock the engineFile rather than uuid-level lock. I think the mutex in engineFile is easy to understand

Little-Wallace: PTAL

Little-Wallace:
I just refactor `ResetEngine`

lightning/backend/local.go
```diff
@@ -1245,7 +1260,6 @@ func (local *local) CleanupEngine(ctx context.Context, engineUUID uuid.UUID) err
 		if err != nil {
 			return err
 		}
-		local.engines.Delete(engineUUID)
```
lance6716, line 1178: this line could be kept since `ResetEngine` no longer calling `CleanupEngine`

Little-Wallace: done

lightning/backend/local.go
```diff
@@ -305,6 +334,26 @@ func NewLocalBackend(
 	return MakeBackend(local), nil
 }
 
+// lockForImport locks the local file for importing.
+func (local *local) lockForImport(engineId uuid.UUID, state importMutexState) bool {
```
glorv, line 338: I think it's better to rename this to `lock` and change the function comment

lightning/backend/local.go
```diff
@@ -1639,7 +1728,10 @@ func (w *LocalWriter) flushKVs() error {
 		return err
 	}
 	w.writeBatch = w.writeBatch[:0]
-	return w.local.db.Ingest([]string{filePath})
+	w.local.lockForImport(importMutexStateNoLock)
```
glorv, line 1731:
```suggestion
	w.local.lockForImport(importMutexStateFlush)
```

lightning/backend/local.go
```diff
@@ -1608,18 +1692,23 @@ func (w *LocalWriter) writeRowsLoop() {
 		totalSize += batchSize
 		log.L().Info("write data by sort index", zap.Int64("bytes", totalSize))
 	}
+	w.local.lockForImport(importMutexStateNoLock)
```
glorv, line 1695:
```suggestion
	w.local.lockForImport(importMutexStateIngest)
```
Since we define enum for each lock action, should avoid using the no-lock when actually do lock

lightning/backend/local.go
```diff
@@ -305,6 +334,26 @@ func NewLocalBackend(
...
+	if e, ok := local.engines.Load(engineId); ok {
+		engine := e.(*LocalFile)
+		engine.lockForImport(state)
+		return true
+	}
+	return false
+}
+
+// unlockForImport unlocks the local file from importing.
+func (local *local) unlockForImport(engineId uuid.UUID) bool {
```
glorv, line 348: ditto

lightning/backend/local.go
glorv: I think it's ok to only depend on the LocalFile level lock after changing the reset engine operation inplace.

lightning/backend/local.go
```diff
@@ -154,6 +168,21 @@ func (e *LocalFile) getSizeProperties() (*sizeProperties, error) {
 	return sizeProps, nil
 }
 
+func (e *LocalFile) isLocked() bool {
```
glorv, line 171: This function is for disk-quota usage right? But I think disk-quota need not to check  isLocked, but isLockedForImport, since if it's locked but not during importing, disk quota should still wait the lock and do import after available. @kennytm  am I right?

kennytm: Let's keep this like this for now, we can add `isLockedForImport` in the disk quota PR.

Little-Wallace: Yes. It is only for disk-quota usage

lightning/backend/local.go
Little-Wallace: But which value shall we set for `unlock`?

Little-Wallace: If we set it to `importMutexStateIngest ` when calling `unlock`, it seems too strange....

lightning/backend/local.go
Little-Wallace: I will remove this method and you can add what you want.

lightning/backend/local.go
glorv: Unlock should always reset the value to `importMutexStateNoLock`, if most parameter for `lockForImport` is `importMutexStateNoLock`, this enum makes little sense, just replace it with `lock` (only lock but not set the enum), and `lockForImport` (lock and set importMutexStateImporting) is clearer.

glorv:
LGTM

ti-srebot (approved):

kennytm:
/merge

```
[2021-01-07T07:30:06.979Z] dial tcp 127.0.0.1:4000: connect: connection refused
```

Issue #550, opened 2021-01-14
panic in tidb backend in strict sql-mode


glorv:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do? If possible, provide a recipe for reproducing the error.
Use lightning tidb backend  to import data with config:
```
[tidb]
sql-mode = "STRICT_ALL_TABLES"
```
panic backtrace:
```
goroutine 578 [running]:
github.com/pingcap/tidb/types.(*Datum).ConvertTo(0xc00a0fd220, 0xc0001dcdc0, 0xc0004a0be8, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
  /go/pkg/mod/github.com/pingcap/tidb@v1.1.0-beta.0.20201126095401-55c106afb89b/types/datum.go:843 +0xcff
github.com/pingcap/tidb/table.CastValue(0x2b48760, 0xc000c6e000, 0x5, 0x0, 0x25442e2, 0xb, 0xc00cb05371, 0x8, 0x8, 0x0, ...)
  /go/pkg/mod/github.com/pingcap/tidb@v1.1.0-beta.0.20201126095401-55c106afb89b/table/column.go:244 +0xf2
github.com/pingcap/tidb-lightning/lightning/backend.(*tidbEncoder).appendSQL(0xc00810e040, 0xc000c22120, 0xc00a0fd660, 0xc008097770, 0x0, 0x0)
  /home/jenkins/agent/workspace/optimization-build-tidb-linux-amd/go/src/github.com/pingcap/tidb-lightning/lightning/backend/tidb.go:181 +0x594
github.com/pingcap/tidb-lightning/lightning/backend.(*tidbEncoder).Encode(0xc00810e040, 0xc019f0a3c0, 0xc00062a480, 0xa, 0x10, 0x1, 0xc00a0ec060, 0xa, 0xb, 0x1f37685, ...)
  /home/jenkins/agent/workspace/optimization-build-tidb-linux-amd/go/src/github.com/pingcap/tidb-lightning/lightning/backend/tidb.go:251 +0x32d
github.com/pingcap/tidb-lightning/lightning/restore.(*chunkRestore).encodeLoop(0xc000c46040, 0x2b02ec0, 0xc008097680, 0xc019f0a360, 0xc008063aa0, 0xc019f0a3c0, 0x2adba00, 0xc00810e040, 0xc00a0ec000, 0xc0192fc000, ...)
[...]

PR #378, opened 2020-08-25
encoder: check string value for tidb encoder

glorv:
### What problem does this PR solve? 
close #205

### What is changed and how it works?
if set sql-mode to strict mode, will check each string value by `table.CastValue`

### Check List 

Tests 

 - Unit test
 - Integration test
 - Manual test (add detailed scripts or steps below)
 - No code

Side effects

Related changes

PR #551, opened 2021-01-15
backend/tidb: temporarily disable the strict-mode value check in tidb backend 

glorv:
#378 added string value check in tidb backend if the SQL mode is strict. But this causes #550  because tidb backend doesn't fetch the full table meta which causes CastValue panic. 

Disable this feature before we come up with a correct solution.

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Side effects

Related changes

### DIFF
--- a/lightning/backend/tidb.go
+++ b/lightning/backend/tidb.go
@@ -188,13 +188,13 @@ func (enc *tidbEncoder) appendSQL(sb *strings.Builder, datum *types.Datum, col *
 		value := strconv.AppendFloat(buffer[:0], datum.GetFloat64(), 'g', -1, 64)
 		sb.Write(value)
 	case types.KindString:
-		if enc.mode.HasStrictMode() {
-			d, err := table.CastValue(enc.se, *datum, col.ToInfo(), false, false)
-			if err != nil {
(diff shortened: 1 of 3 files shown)


lightning/backend/tidb.go
```diff
@@ -188,13 +188,13 @@ func (enc *tidbEncoder) appendSQL(sb *strings.Builder, datum *types.Datum, col *
...
 		sb.Write(value)
 	case types.KindString:
-		if enc.mode.HasStrictMode() {
-			d, err := table.CastValue(enc.se, *datum, col.ToInfo(), false, false)
-			if err != nil {
-				return errors.Trace(err)
-			}
-			datum = &d
-		}
+		//if enc.mode.HasStrictMode() {
```
kennytm, line 191: please add a link to #550.

kennytm (commented):
Rest LGTM

lightning/backend/tidb.go
glorv: done

kennytm:
LGTM

ti-srebot (approved):

